PR #2, opened 2020-10-13
Goinit

sebastianst:
Adds `tee` package with `Enclave` interface and core types.


tee/tee.go
```diff
@@ -0,0 +1,71 @@
...
+		// GenKeys generates a new secp256k1 ECDSA key and stores it as the
+		// enclave's signing key.
+		//
+		// It returns the public key derived Ethereum address and attestation of
+		// correct initialization of the enclave with the generated address. The
+		// attestation can be used to verify the enclave with the TEE vendor.
+		//
+		// The Operator must deploy the contract with the Enclave's address after
+		// calling GenKeys.
+		GenKeys() (common.Address, []byte, error)
```
RmbRT: I think that GenKey() or Setup() or Init() or something like that would be better here (as only one key is generated). I would also name the returned values OR make a type alias for the attestation.

tee/tee.go
```diff
@@ -0,0 +1,71 @@
...
+	Epoch = uint64
+
+	// A Block is a go-ethereum block together with its receipts. go-ethereum's
+	// types.Block type doesn't store the receipts...
+	Block struct {
+		types.Block
+		Receipts types.Receipts
+	}
+
+	Transaction struct {
```
RmbRT, line 57: Doesn't this need a sender and recipient?

tee/tee.go
```diff
@@ -0,0 +1,71 @@
...
+		epoch   Epoch
+		account common.Address
+		amount  *big.Int
+	}
+
+	Proof struct {
```
RmbRT: I think BalanceProof is clearer, but not absolutely necessary.

tee/tee.go
RmbRT: and what about user signatures?

RmbRT (changes requested):

tee/tee.go
ndzik: Having it called `Init()` would also be more in line with other SGX implementations I saw.

sebastianst: Guys, this is a hackathon, lower your standards ;) Renamed to `Init`.

tee/tee.go
sebastianst: You're totally right :joy:

PR #3, opened 2020-10-14
Go Prototype Enclave

sebastianst:


tee/prototype/txcache.go
```diff
@@ -0,0 +1,35 @@
...
+
+type txCache struct {
+	senders    map[common.Address][]*tee.Transaction
+	recipients map[common.Address][]*tee.Transaction
+	txs        map[common.Hash]*tee.Transaction
+}
+
+// TODO: Pointer receiver?
+// cacheTx caches the given transaction.
+func cacheTx(cache txCache, tx *tee.Transaction) txCache {
```
sebastianst: why isn't this just a method on the `txCache`?

tee/prototype/epochproc.go
```diff
@@ -0,0 +1,397 @@
...
+package prototype
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	err "perun.network/go-perun/pkg/errors"
```
sebastianst: I wouldn't use `err`, it's the usual error variable name. Could lead to unintended shadowing. Bette like `perrors`.

tee/prototype/epochproc.go
```diff
@@ -0,0 +1,397 @@
...
+			continue
+		}
+		for _, tx := range sTxs {
+			delete(txs.txs, tx.Hash())
+		}
+		for _, tx := range rTxs {
+			delete(txs.txs, tx.Hash())
+		}
+	}
+	prunedTxs := make([]*tee.Transaction, 0, len(txs.txs))
```
sebastianst:
```suggestion
	prunedTxs := make([]*tee.Transaction, len(txs.txs))
```
or the loop will panic

tee/prototype/epochproc.go
```diff
@@ -0,0 +1,397 @@
...
+	i := 0
```
sebastianst: you forgot to increase the counter

sebastianst: or use `append`.

tee/prototype/epochproc.go
```diff
@@ -0,0 +1,397 @@
...
+		case tx := <-txs:
+			stagedTxs = cacheTx(stagedTxs, tx)
+		}
+	}
+}
+
+// getBalProofs retrieves a slice of `*tee.BalanceProof`s from the balance proof cache.
+func getBalProofs(bps balProofCache) []*tee.BalanceProof {
+	proofs := make([]*tee.BalanceProof, len(bps))
+	i := 0
```
sebastianst: increase counter in loop

tee/prototype/epochproc.go
```diff
@@ -0,0 +1,397 @@
...
+			phaseDone <- exiters
+			close(phaseDone) // phase done, stop TX processor.
+			return nil
+		}
+	}
+	return nil
+}
+
+// retrieveCachedDepProofs reduces the deposit proof cache to a slice of `tee.DepositProof`s.
+func retrieveCachedDepProofs(cache map[common.Address]*tee.DepositProof) []*tee.DepositProof {
```
sebastianst: This func and `getBalProofs` do conceptually the same, maybe we call them similarly? like `asFooProofs`? Also, in one function you use `append` in the other you address the slice index.

tee/prototype/epochproc.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	return &tee.DepositProof{
+		Balance: b,
+		Sig:     sig,
+	}, nil
+}
+
+// updateBalanceProofs updates the balance proof cache with for given transaction.
+func (e *Enclave) updateBalanceProofs(bpc balProofCache, ep *Epoch, sender, recipient common.Address) (balProofCache, error) {
```
sebastianst: a map is a pointer type, no need to return it, you work on the same instance as the caller.

tee/prototype/epochproc.go
```diff
@@ -0,0 +1,397 @@
...
+		}
+		if accBal, ok := ep.balances[deposit.Account]; ok {
+			accBal.Value.Add(accBal.Value, deposit.Value)
+		} else {
+			ep.balances[deposit.Account] = Bal{
+				0,
+				new(big.Int).Set(deposit.Value),
+			}
+		}
+		if deposit.Epoch != ep.Number {
```
sebastianst: we should do this check before we do the balance updates.

tee/prototype/epochproc.go
```diff
@@ -0,0 +1,397 @@
...
+		accBal.SetUint64(0)
+		delete(ep.balances, exit.Account)
+		exiters = append(exiters, exit.Account)
+	}
+
+	return exiters, nil
+}
+
+// applyEpochTx adjusts `e.balances` according to given transactions.
+func (e *Enclave) applyEpochTx(ep *Epoch, txs []*tee.Transaction, bpc balProofCache) (balProofCache, error) {
```
sebastianst: why return the cache? also note that in the last line, you just return `nil, nil`.

tee/prototype/events.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/perun-network/erdstall/contracts/bindings"
+)
+
+var depositedEvent common.Hash = crypto.Keccak256Hash([]byte("Deposited(uint64,address,uint256)"))
+
+var exitingEvent common.Hash = crypto.Keccak256Hash([]byte("Exiting(uint64,address,uint256)"))
```
sebastianst, line 19: can't we extract those from the generated bindings?

tee/prototype/epochproc.go
```diff
@@ -0,0 +1,397 @@
...
+			Account: acc,
+			Value:   bal.Value,
+		}
+
+		bp, err := e.signBalanceProof(b)
+		if err != nil {
+			return nil, fmt.Errorf("generating balance proofs: %w", err)
+		}
+		balProofs[acc] = bp
+		i++
```
sebastianst, line 182: funny, in this loop you do increase the counter, but never use it :joy:

tee/prototype/epochproc.go
ndzik: good point.

tee/prototype/epochproc.go
sebastianst: or use append with `make(foo, 0, capacity)`.

tee/prototype/epochproc.go
ndzik: of course! Got too late :see_no_evil:

tee/prototype/epochproc.go
ndzik: This is true. The reason I went for this is, because in the usage it's clear that `balProofCache` is being modified within that function. Feels more functional :see_no_evil:  So looking at

``` Go
		bpc, err = e.updateBalanceProofs(bpc, ep, tx.Sender, tx.Recipient)
		if err != nil {
			return nil, fmt.Errorf("applying tx in txEpoch %v: %w", ep.Number, err)
		}
```
makes it obvious that bpc gets changed. What do you think?

tee/tee.go
```diff
@@ -79,6 +84,7 @@ type (
 		Recipient common.Address
 		Amount    *big.Int
 		Sig       []byte
+		hash      common.Hash
```
sebastianst, line 87: you use this as a map key, why does this value also need to be in the struct?

tee/prototype/epochproc.go
ndzik: see [comment](https://github.com/perun-network/erdstall/pull/3#discussion_r507686815).

and good catch.

tee/prototype/epochproc.go
sebastianst: If we directly apply, instead of cache, we don't use this anymore, right?

tee/tee.go
ndzik: It made sense to me since it is tied to the transaction itself and I thought there might be other use-cases. This also allowed me to cache the hash instead of recomputing it every time when its needed. But I am okay if we remove this.

sebastianst: Ah ok, if you call `Hash()` several times and cache the calls, it's fine.

tee/prototype/epochproc.go
ndzik: Got rid of it. :ok_hand:

sebastianst (commented):
I'd say let's give it a try with the operator.

PR #6, opened 2020-10-15
Add parameters, proof and tx encoding and sig verification

sebastianst:


contracts/contracts/Erdstall.sol
```diff
@@ -182,6 +182,15 @@ contract Erdstall {
...
     function verifyBalance(Balance memory balance, bytes memory sig) internal view {
-        require(Sig.verify(abi.encode(balance), sig, tee), "invalid signature");
+        require(Sig.verify(encodeBalanceProof(balance), sig, tee), "invalid signature");
+    }
+
+    function encodeBalanceProof(Balance memory balance) internal view returns (bytes memory) {
+        return abi.encode(
+            "ErdstallBalance",
+            address(this),
+            balance.epoch,
```
ggwpez, line 192: Just pass `balance` instead of the single members, just like in the go code.

tee/params.go
```diff
@@ -0,0 +1,43 @@
...
+
+type Parameters struct {
+	PowDepth         uint64         // required confirmed block depth
+	PhaseDuration    uint64         // number of blocks of one phase (not epoch)
+	ResponseDuration uint64         // challenge response grace period for operator at end of exit phase
+	InitBlock        uint64         // block at which Erdstall contract was deployed
+	TEE              common.Address // Enclave's public key address
+	Contract         common.Address // Erdstall contract address
+}
+
```
ggwpez, line 15: For completeness, we should add a constructor that checks the requirements as in the contract.  
Currently only `require(2 * _responseDuration <= _phaseDuration, "responseDuration too long");`.  
But not doing it is also ok, since then the deployment just fails.

sebastianst: Yes good point, but I think it's fine if it fails during deployment. The clients would just read those parameters from the deployed contract anyways, where they are guaranteed to be correct.

ggwpez: We discussed to let the deployment fail.

ggwpez (approved):

PR #7, opened 2020-10-15
Deposit recovery

sebastianst:
I was missing the deposit recovery in case someone funded but never received any balance proof.


matthiasgeihs (commented):
lgtm

PR #9, opened 2020-10-16
Operator

matthiasgeihs:


matthiasgeihs:
native error wrapping should be used
`fmt.Errorf("doing foo: %w", err)`
https://blog.golang.org/go1.13-errors

cmd/operator/enclave.go
```diff
@@ -0,0 +1,119 @@
...
+	return &RemoteEnclave{operator: operator}
+}
+
+// AddTransaction adds a transaction to the enclave's transaction pool.
+func (e *RemoteEnclave) AddTransaction(t tee.Transaction, _ struct{}) error {
+	return e.Enclave.ProcessTXs(&t)
+}
+
+// GetDepositProof returns the next deposit proof for the given user.
+func (e *RemoteEnclave) GetDepositProof(user common.Address, _dp *tee.DepositProof) error {
```
sebastianst: Can we somehow make a subscription on the client's side? That would be much better than having the client poll the operator for proofs.

cmd/operator/enclave.go
```diff
@@ -0,0 +1,119 @@
...
+		return errors.New("deposit proof not available")
+	}
+
+	*_dp = *dp
+
+	return nil
+}
+
+// GetBalanceProof returns the next balance proof for the given user.
+func (e *RemoteEnclave) GetBalanceProof(user common.Address, _bp *tee.BalanceProof) error {
```
sebastianst: See other comment, sub on client side possible?

cmd/operator/operator.go
```diff
@@ -0,0 +1,92 @@
...
+	}()
+
+	// Handle deposit proofs
+	go func() {
+		for {
+			dps, err := operator.Enclave.DepositProof()
+			if err != nil {
+				log.Fatal("failed to retrieve deposit proofs", err)
+			}
+			operator.depositProofs.AddAll(dps)
```
sebastianst: See other comment, if we can have rpc subscriptions, we can push here.

cmd/operator/operator.go
```diff
@@ -0,0 +1,92 @@
...
+	}()
+
+	// Handle balance proofs
+	go func() {
+		for {
+			bps, err := operator.Enclave.BalanceProofs()
+			if err != nil {
+				log.Fatal("failed to retrieve balance proofs", err)
+			}
+			operator.balanceProofs.AddAll(bps)
```
sebastianst: See other comment, if we can have rpc subscriptions, we can push here.

cmd/operator/enclave.go
matthiasgeihs: possible, but we decided to go for plain rpc for now. might change if we run into problems with that approach.

cmd/operator/enclave.go
matthiasgeihs: see above

cmd/operator/operator.go
matthiasgeihs: see above

cmd/operator/operator.go
matthiasgeihs: see above

cmd/operator/test/user.go
```diff
@@ -0,0 +1,80 @@
...
+	defer cancel()
+
+	t, err := u.ethClient.NewKeyedTransactor(ctx, gasLimit)
+	if err != nil {
+		u.Fatal("creating keyed transactor:", err)
+	}
+
+	t.Value = new(big.Int).SetInt64(amount)
+
+	u.contract.Deposit(t)
```
sebastianst: Depending on how you want to use this, you might want to also call `bind.WaitMined` after you send the tx.

cmd/operator/operator.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	go func() {
+		defer blockSub.Unsubscribe()
+		for {
+			select {
+			case b := <-blockSub.Blocks():
+				operator.enclave.ProcessBlocks(b)
+				log.Println("processed new block")
+			case <-shutdown:
+				break
```
ndzik: I am not sure if this really exits the `for` loop. [Go's language specification](https://golang.org/ref/spec#Break_statements) says about `break` statements:

> A "break" statement terminates execution of the innermost "for", "switch", or "select" statement within the same function.

cmd/operator/test/user.go
matthiasgeihs: good point

cmd/operator/operator.go
matthiasgeihs: ok, not sure. wasn't using the shutdown functionality anyhow. so i removed it for now.

tee/prototype/enclave.go
```diff
@@ -59,6 +58,17 @@ func NewEnclave(wallet accounts.Wallet) *Enclave {
 	}
 }
 
+func NewEnclaveWithAccount(wallet accounts.Wallet, account accounts.Account) *Enclave {
+	return &Enclave{
```
sebastianst: The latest version or `NewEnclave` looks different, better don't copy the code from `NewEnclave` but call `NewEnclave` and then set the account field manually. Otherwise we have to keep copied code in sync.

sebastianst (changes requested):

tee/prototype/enclave.go
matthiasgeihs: ok, done

PR #13, opened 2020-10-19
Expand ETH client.

ggwpez:
In order to reuse our existing code, i changed the ContractInterface
to a ContractBackend, then we can use our Transactor creation functions
and do not need new NewWatchOpts and NewFilterOpts functions.

### DIFF
--- a/eth/client.go
+++ b/eth/client.go
@@ -20,13 +20,15 @@ import (
 	"github.com/perun-network/erdstall/tee"
 )
 
-const defaultGasLimit = 2000000
+const (
+	defaultGasLimit = 2000000
+	defaultPowDepth = 0
+)
 
 type (
 	// Client is an Ethereum client to interact with the Erdstall contract.
 	Client struct {
-		peruneth.ContractInterface
-		tr      peruneth.Transactor
+		peruneth.ContractBackend
 		account accounts.Account
 	}
 
@@ -39,11 +41,10 @@ type (
 
 // NewClient creates a new Erdstall Ethereum client.
 func NewClient(
-	ci peruneth.ContractInterface,
-	tr peruneth.Transactor,
+	cb peruneth.ContractBackend,
 	a accounts.Account,
 ) *Client {
-	return &Client{ci, tr, a}
+	return &Client{cb, a}
 }
 
 // NewClientForWallet returns a new Client using the given wallet as transactor.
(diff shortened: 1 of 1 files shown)


eth/client.go
```diff
@@ -37,13 +36,17 @@ type (
 	}
 )
 
+const (
+	defaultPowDepth = uint64(6)
+	gasLimit        = uint64(6000000)
```
sebastianst: There already is a `defaultGasLimit` in line 23. Please use this one and adjust if necessary. It's confusing to have two consts for that.

eth/client.go
```diff
@@ -37,13 +36,17 @@ type (
 	}
 )
 
+const (
+	defaultPowDepth = uint64(6)
```
sebastianst: Please set to 0. You also don't necessarily need to set a const number type, afaik.

eth/client.go
```diff
@@ -108,6 +112,42 @@ func (cl *Client) SubscribeToBlocks() (*BlockSubscription, error) {
 	return &BlockSubscription{sub, blocks}, nil
 }
 
+// SubscribeToDeposited writes past Deposited events and newly received ones
+// into the sink. Can be cancelled via context.
+func (cl *Client) SubscribeToDeposited(ctx context.Context, contract *bindings.Erdstall, epochs []uint64, sink chan *bindings.ErdstallDeposited) error {
```
sebastianst: So `epochs` is optional to only get specific epochs?

eth/client.go
```diff
@@ -169,9 +183,9 @@ func (cl *Client) DeployContracts(params *tee.Parameters) error {
 	ctx, cancel := NewDefaultContext()
 	defer cancel()
 
-	tr, err := cl.NewTransactor(ctx, defaultGasLimit)
+	tr, err := cl.ContractBackend.NewTransactor(ctx, big.NewInt(0), gasLimit, cl.account)
```
sebastianst: use `defaultGasLimit`.

eth/client.go
```diff
@@ -198,6 +212,38 @@ func (cl *Client) DeployContracts(params *tee.Parameters) error {
 	return nil
 }
 
+func (cl *Client) BindContract(ctx context.Context, addr common.Address) (*tee.Parameters, *bindings.Erdstall, error) {
```
sebastianst, line 215: Nice :sparkles:

sebastianst (changes requested):

eth/client.go
ggwpez: Oh fail i did not see that one. Will move the other variable up.

eth/client.go
ggwpez: As i understand it, it is not optional. go-ethereum will only return deposited events with the specified epochs. I amended the comment.

sebastianst: I looked it up, it is optional

sebastianst (approved):

PR #19, opened 2020-10-21
Rpc

RmbRT:
Created RPC node and RPC connection types for connecting the TEE with the outside world.


operator/operator.go
```diff
@@ -24,9 +24,9 @@ import (
 
 // Operator resprents a TEE Plasma operator.
 type Operator struct {
-	enclave   tee.Enclave
-	params    tee.Parameters
-	ethClient *eth.Client
+	deposit, balance, blocks tee.Enclave // different handles to same Enclave.
```
matthiasgeihs: This looks odd. Why do you use multiple handles to the same enclave?

operator/operator.go
```diff
@@ -55,7 +55,33 @@ func New(
...
+		params:            params,
+		ethClient:         client,
+		depositProofs:     newDepositProofs(),
+		balanceProofs:     newBalanceProofs(),
+		contract:          _contract,
+		respondChallenges: respondChallenges,
+	}, nil
+}
+
+func NewRemote(
```
matthiasgeihs: Please add documentation, at least a short one.

What is `NewRemote`? Does it generate a *remote* operator? If yes, what is a remote operator? Isn't an operator always operating remotely? Consider renaming or adding comments, otherwise this is not clear.

operator/operator.go
```diff
@@ -120,7 +167,7 @@ func (operator *Operator) Serve(port int) error {
 	}
 
 	// Start enclave
-	errGo("Enclave.Run", func() error { return operator.enclave.Run(operator.params) })
+	errGo("Enclave.Run", func() error { return operator.blocks.Run(operator.params) })
```
matthiasgeihs: You start the whole enclave by calling `Run` on the `blocks` enclave? Confusing.

tee/rpc/operator.go
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package rpc
+
+import (
+	"github.com/perun-network/erdstall/operator"
+)
+
+// Setup creates an operator setup that dials the specified remote enclave.
+func Setup(
```
matthiasgeihs: I think this method should be in the operator package.

operator/operator.go
```diff
@@ -76,6 +102,21 @@ func Setup(cfg *Config) *Operator {
 	log.Debug("Operator.Setup: Enclave account loaded")
 
 	enclave := prototype.NewEnclaveWithAccount(wallet, enclaveAccount)
+
+	return RemoteSetup(enclave, enclave, enclave, cfg)
+}
+
+// RemoteSetup creates an operator from the given configuration and remote
+// enclave. The three enclave handles must point to the same remote enclave.
+func RemoteSetup(
```
matthiasgeihs: Please extend the operator tests to run with `Setup` and `RemoteSetup`.

matthiasgeihs (changes requested):
Having three handles to the same enclave is confusing. How do you justify that separation of different functionalities of the same enclave?

As mentioned before, I don't like that you are implementing your custom RPC functionality instead of using golang's net/rpc package. The code base gets bloated, more error prone, and on top of that it seems like you can get rid of the enclave handle triplet because net/rpc probably supports concurrent requests, which seems to be what you really want here.

**I strongly recommend revising your code to use net/rpc.**

operator/operator.go
RmbRT: restructured

tee/rpc/conn.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// NewRemoteEnclave communicates with a rpc.Node via conn.
+func NewRemoteEnclave(conn io.ReadWriteCloser) *RemoteEnclave {
+	return &RemoteEnclave{
+		client: rpc.NewClient(conn),
+	}
+}
+
+// Dials an enclave three times to allow three concurrent accesses to the enclave.
```
sebastianst: What's this "three times"?

tee/rpc/node.go
```diff
@@ -0,0 +1,127 @@
...
+		log.Panic("Node.Start(): already running")
+	}
+	n.stopped.OnCloseAlways(func() { l.Close() })
+	if err := n.server.Register(n); err != nil {
+		log.Panicf("Node.Start() registering server: %v", err)
+	}
+	go n.server.Accept(l)
+}
+
+// Void indicates that a function has no arguments o no return value.
```
sebastianst: "or"

tee/rpc/node.go
```diff
@@ -0,0 +1,127 @@
...
+type Void struct{}
```
sebastianst: Do we need this exported?

tee/rpc/conn.go
```diff
@@ -0,0 +1,93 @@
...
+func (re *RemoteEnclave) ProcessBlocks(blocks ...*tee.Block) error {
+	encodedBlocks := make([][]byte, len(blocks))
+	for i, b := range blocks {
+		var buf bytes.Buffer
+		if err := b.EncodeRLP(&buf); err != nil {
+			return fmt.Errorf("encoding block: %w", err)
+		}
+		if err := gob.NewEncoder(&buf).Encode(&b.Receipts); err != nil {
+			return fmt.Errorf("encoding receipts: %w", err)
+		}
```
sebastianst: Please encapsulate the encoding of a `tee.Block` into its own method on `tee.Block` and not do it here. Without having looked at it yet, the same probably applies to decoding. You can even use the [custom gob methods](https://pkg.go.dev/encoding/gob#GobEncoder) for that.

tee/rpc/conn.go
```diff
@@ -0,0 +1,93 @@
...
+		encodedBlocks[i] = buf.Bytes()
+	}
+	return re.client.Call("Node.ProcessBlocks", encodedBlocks, &Void{})
```
sebastianst: processing of blocks and transactions now returns an error, why is the result type void?

tee/rpc/conn.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (re *RemoteEnclave) ProcessTXs(txs ...*tee.Transaction) error {
+	return re.client.Call("Node.ProcessTXs", &txs, &Void{})
```
sebastianst: processing of blocks and transactions now returns an error, why is the result type void?

tee/rpc/conn.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (re *RemoteEnclave) Shutdown() {
+	if err := re.client.Call("Node.Shutdown", Void{}, &Void{}); err != nil {
+		log.Error(err)
+	}
+}
+
+func (re *RemoteEnclave) Init() (common.Address, []byte, error) {
+	var res TeeInitRes
+	return res.Addr, res.Sig, re.client.Call("Node.Init", &Void{}, &res)
```
sebastianst: Sure that this works in 1 line? If it reads the values of `res.Addr` and `res.Sig` first and then performs the call, it won't.

tee/rpc/conn.go
sebastianst: or does gob rpc work in such a way that errors are returned as errors of the call?

tee/rpc/node.go
```diff
@@ -0,0 +1,127 @@
...
+	for i, be := range encodedBlocks {
+		r := bytes.NewReader(be)
+		stream := rlp.NewStream(r, 0)
+		var b tee.Block
+		if err := b.DecodeRLP(stream); err != nil {
+			return fmt.Errorf("decoding block: %w", err)
+		}
+		if err := gob.NewDecoder(r).Decode(&b.Receipts); err != nil {
+			return fmt.Errorf("decoding receipts: %w", err)
+		}
```
sebastianst: see other comment, encapsulate as `GobDecode`

tee/rpc/node_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+type mockListener struct{ conn chan net.Conn }
+
+func (l *mockListener) Accept() (net.Conn, error) {
+	c := <-l.conn
+	if c == nil {
+		return nil, errors.New("closed")
+	}
+	return c, nil
+}
```
sebastianst: newline after method spanning several blocks.

tee/rpc/node_test.go
```diff
@@ -0,0 +1,81 @@
...
+func (l *mockListener) dial() (_ net.Conn, err error) {
+	a, b := net.Pipe()
+	err = errors.New("closed")
+	func() {
+		defer func() { _ = recover() }()
+		l.conn <- b
+		err = nil
+	}()
+	return a, err
+}
```
sebastianst: newline after method spanning several blocks.

tee/rpc/node_test.go
```diff
@@ -0,0 +1,81 @@
...
+func (l *mockListener) Accept() (net.Conn, error) {
+	c := <-l.conn
+	if c == nil {
+		return nil, errors.New("closed")
+	}
+	return c, nil
+}
+func (l *mockListener) Close() (_ error) { close(l.conn); return }
+func (*mockListener) Addr() net.Addr     { panic(nil) }
+func (l *mockListener) dial() (_ net.Conn, err error) {
```
sebastianst: newline before method spanning several blocks.

tee/rpc/node_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	test.AssertTerminates(t, 10*time.Second, func() {
+		_, _, err = enc.Init()
+		assert.NoError(t, err)
+		assert.NoError(t, enc.Run(tee.Parameters{}))
+		assert.NoError(t, enc.ProcessBlocks())
+		assert.NoError(t, enc.ProcessTXs())
+		_, err = enc.DepositProofs()
+		assert.NoError(t, err)
+		_, err = enc.BalanceProofs()
```
sebastianst: you don't test that the correct data is actually passed around, is this part of another test?

tee/rpc/main.go
```diff
@@ -0,0 +1,47 @@
...
+
+	hdwallet "github.com/miguelmota/go-ethereum-hdwallet"
+	"github.com/perun-network/erdstall/tee/prototype"
+
+	"perun.network/go-perun/log"
+)
+
+// main(listenAddr:port, mnemonic, derivation path)
+// nolint:unused,deadcode
+func main() {
```
sebastianst: I think the main belongs into its own package in `cmd` like `cmd/renclave`

tee/rpc/node.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/perun-network/erdstall/tee"
+
+	"perun.network/go-perun/log"
+	"perun.network/go-perun/pkg/sync"
+	"perun.network/go-perun/pkg/sync/atomic"
+)
+
+// Node is a server that exposes an Enclave as a slave.
+type Node struct {
```
sebastianst: Not sure if `Node` is the best name for this. A node is usually one entity in a network, like an ethereum node. This here is rather a server, where the client is the `RemoteEnclave`. So maybe we call this something like `RPCEnclave`. We should also sync with @ggwpez since he's having similar naming issues in the operator in #61

operator/operator.go
```diff
@@ -67,15 +67,23 @@ func New(
...
-	enclaveAccount, err := wallet.Derive(enclaveAccountDerivationPath, true)
-	AssertNoError(err)
-	log.Debug("Operator.Setup: Enclave account loaded")
+	if enclave == nil {
+		enclaveAccountDerivationPath := hdwallet.MustParseDerivationPath(cfg.EnclaveDerivationPath)
+		enclaveAccount, err := wallet.Derive(enclaveAccountDerivationPath, true)
+		AssertNoError(err)
+		log.Debug("Operator.Setup: Enclave account loaded")
+		enclave = prototype.NewEnclaveWithAccount(wallet, enclaveAccount)
+	}
```
sebastianst: better move this code above into `Setup` and don't allow `nil` enclaves.

operator/operator.go
```diff
@@ -67,15 +67,23 @@ func New(
 
 // Setup creates an operator from the given configuration.
 func Setup(cfg *Config) *Operator {
```
sebastianst: I'd rename this to `SetupWithPrototypeEnclave` and `SetupWithEnclave` to `Setup`.

operator/operator.go
```diff
@@ -119,6 +120,20 @@ func SetupWithEnclave(cfg *Config, enclave tee.Enclave) *Operator {
 	return operator
 }
 
+// RemoteSetup creates an operator setup that dials the specified remote
+// enclave.
+func RemoteSetup(
```
sebastianst: And to be consistent, then call this `SetupWithRemoteEnclave`

operator/operator_test.go
```diff
@@ -136,7 +141,9 @@ func initEnvironment(t *testing.T) *environment {
 		log.Fatal(err)
 	}
 
-	operator := Setup(cfg)
+	enclave := createRemoteEnclave(cfg)
+	operator := SetupWithEnclave(cfg, enclave)
```
sebastianst: Can we instead parameterize the test so that it can be run with a remote as well as a prototype enclave.
This test here should test the operator and not be a remote enclave test, so I think it would be better to have this test still use the prototype.

operator/operator_test.go
```diff
@@ -230,3 +257,28 @@ func newDefaultConfig() *Config {
 		true,
 	}
 }
+
+type mockListener struct{ conn chan net.Conn }
+
+func (l *mockListener) Accept() (net.Conn, error) {
```
sebastianst: add newlines before and after method definitions spanning several lines

sebastianst (changes requested):

tee/rpc/node.go
RmbRT: Sadly, yes. Go RPC does not allow unexported types in arguments of service functions.

tee/rpc/conn.go
RmbRT: Outdated, good find.

tee/rpc/conn.go
RmbRT: Errors are returned as the error return value of the service function.

tee/rpc/conn.go
RmbRT: Doing a custom Gob encoder sounds like overkill.

tee/rpc/conn.go
RmbRT: Exactly.

tee/rpc/node.go
RmbRT: I'm not exactly sure how I would use a GobDecoder, I'd rather just use our Encoder/Decoder interface.

tee/rpc/node_test.go
RmbRT: This is tested implicitly in the operator test that uses a remote enclave. If the data cannot be decoded, net/rpc will not forward the call. This means we already know that the encoding and decoding worked. And it's quite trivial to see that the encoding and decoding methods contain the same logic.
Should I still write another separate test just for this?

operator/operator_test.go
RmbRT: Added parameterization options for prototype enclave, remote enclave (dial), and remote enclave (wrapped around local prototype enclave).

tee/rpc/rpcenclave.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (re *RPCEnclave) ProcessBlocks(blocks ...*tee.Block) error {
+	enc := make([][]byte, len(blocks))
+	for i, b := range blocks {
+		var buf bytes.Buffer
+		if err := b.Encode(&buf); err != nil {
+			return fmt.Errorf("Encoding block %d: %w", i, err)
+		}
+		enc[i] = buf.Bytes()
```
sebastianst: Referring to your answer that a custom `GobEncode` is overkill. I think it's literally the same amount of code that you have now, but better encapsulated and cleaner. If you change `tee.Block.Encode` to `GobEncode() ([]byte, error)` ([`GobEncoder`](https://pkg.go.dev/encoding/gob#GobEncoder)), then I think this whole function here just collapses into a single line
```go
 	return re.client.Call("Server.ProcessBlocks", blocks, &Void{})
```
It basically just moves the `var buf bytes.Buffer` and `buf.Bytes()` bits into the encoder function.

Something similar is probably true for decoding.

cmd/renclave/main.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/perun-network/erdstall/tee/rpc"
+
+	"perun.network/go-perun/log"
+)
+
+// main(listenAddr:port, mnemonic, derivation path)
+// nolint:unused,deadcode
+func main() {
+	if len(os.Args) != 4 {
+		log.Fatalf("main: expected <listen addr> <mnemonic> <derivation path>")
```
sebastianst:
```suggestion
		log.Fatalf("Usage: %s <listen addr> <mnemonic> <derivation path>", os.Args[0])
```

operator/operator_test.go
```diff
@@ -164,7 +170,79 @@ func initEnvironment(t *testing.T) *environment {
...
+}
+
+type testSetup interface {
+	setup(cfg *Config) (operator *Operator, cleanup func() error)
+}
+
+type prototypeSetup struct{}
+
+func (prototypeSetup) setup(cfg *Config) (*Operator, func() error) {
+	return SetupWithPrototypeEnclave(cfg), func() error { return nil }
```
sebastianst: why don't you `Stop` the prototype enclave in the cleanup function here?

operator/operator_test.go
```diff
@@ -114,11 +118,12 @@ type environment struct {
 	user2             *test.User
 	errg              *errors.Gatherer
 	enclaveParameters tee.Parameters
+	stopEnclave       func() error
```
sebastianst: Do you ever call `stopEnclave` from the test? I'd also rename this field to `cleanup` because that's how you call the return function of `testSetup.setup`

sebastianst (changes requested):
Instead of testing the remote enclave (the two versions `remoteSetup` and `simRemoteSetup`) with the `operator/operator_test.go` test, change the enclave test in `tee/prototype` to a generic enclave test and then call the generic test with all three versions. Only use the prototype enclave in the operator test, as before.

operator/operator_test.go
RmbRT: reverted all changes to the operator test.

operator/operator_test.go
RmbRT: code no longer exists

tee/test/generictest.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/stretchr/testify/require"
+	"perun.network/go-perun/pkg/test"
+
+	"github.com/perun-network/erdstall/eth"
+	"github.com/perun-network/erdstall/tee"
+)
+
+type clientAction = func(ctx context.Context, bal *tee.BalanceProof) error
+
+type Client interface {
```
sebastianst: We discussed to move `tee/test.NewRandomAddress` into package `eth` to break the dependency and then to delete this interface and use the normal `client/test.Client` in the `GenericEnclaveTest`.

tee/test/generictest.go
```diff
@@ -0,0 +1,268 @@
...
+	Exit(context.Context, *tee.BalanceProof) error
+	Withdraw(context.Context, *tee.BalanceProof) error
+	Send(common.Address, *big.Int) error
+	SendToClient(Client, *big.Int) error
+	SendInvalidTxs(*rand.Rand, common.Address) []error
+	InvalidTxs(*rand.Rand, common.Address) []*tee.Transaction
+	SignTx(*tee.Transaction)
+}
+
+type EnclaveTestConfig interface {
```
sebastianst: `NewClient` won't be necessary anymore, see other comment.

Do we even need `NewTransactor`? Don't we always use the direct enclave transactor in the enclave test? If so, we can also delete `NewTransactor` and always use the `EnclaveTransactor` in the test, so we don't need `EnclaveTestConfig` at all.

tee/tee.go
```diff
@@ -123,6 +124,8 @@ type (
 	Amount big.Int
 )
 
+var ErrEnclaveStopped = errors.New("Enclave stopped")
+
```
sebastianst, line 128: this makes sense to move here :+1:

tee/rpc/rpc_test.go
```diff
@@ -82,3 +87,28 @@ func TestConn(t *testing.T) {
...
+
+type Config struct{}
+
+func (Config) NewTransactor(e tee.Enclave) tee.Transactor {
+	return &test.EnclaveTransactor{Enclave: e}
+}
+
+func (Config) NewClient(p tee.Parameters, w tee.TextSigner, c *eth.Client, t tee.Transactor) (ttest.Client, error) {
+	return test.NewClient(p, w, c, t)
+}
```
sebastianst: See other comments, can probably all be deleted.

sebastianst (changes requested):

tee/test/generictest.go
```diff
@@ -0,0 +1,243 @@
...
+	ctest "github.com/perun-network/erdstall/client/test"
+	"github.com/perun-network/erdstall/eth"
+	"github.com/perun-network/erdstall/tee"
+)
+
+type clientAction = func(ctx context.Context, bal *tee.BalanceProof) error
+
+func GenericEnclaveTest(t *testing.T, enc tee.Enclave) {
+	assert, requiree := assert.New(t), require.New(t)
+	_ = assert
```
sebastianst, line 28: Please boyscout-remove

sebastianst (approved):
Since I only had one trivial comment, I'll just merge it.

PR #22, opened 2020-10-21
Fix operator tests

matthiasgeihs:


eth/client.go
```diff
@@ -155,6 +166,62 @@ func (cl *Client) SubscribeToBlocks() (*BlockSubscription, error) {
...
+
+	headerSub, err := cl.SubscribeNewHead(ctx, headers)
+	if err != nil {
+		return nil, fmt.Errorf("subscribing to blockchain head: %w", err)
+	}
+
+	blockSub := &BlockSubscription2{headerSub, blocks, errChan, startBlockNumber}
+
+	errg := errors.NewGatherer()
+	errg.Go(func() error {
```
sebastianst: I wouldn't use an error gatherer to start a single routine and then start another routine just to write the result in the channel. you could do here e.g.
```suggestion
	subFn := func() error { // function body as it is }
	go func() {
		errChan <- subFn()
	}()
```

matthiasgeihs: should be
```
go func() {
	err := subFn()
	errChan <- err
}()
```

ok, done

eth/client.go
```diff
@@ -155,6 +165,62 @@ func (cl *Client) SubscribeToBlocks() (*BlockSubscription, error) {
...
+	}
+
+	blockSub := &BlockSubscription2{headerSub, blocks, errChan, startBlockNumber}
+
+	run := func() error {
+		for {
+			select {
+			case err := <-headerSub.Err():
+				if err != nil {
+					errChan <- fmt.Errorf("subscription error: %w", err)
```
sebastianst: you write into the err chan here and then return the error, which will write it again?

matthiasgeihs: nice catch

PR #25, opened 2020-10-21
Add Run, Shutdown to tee.Enclave interface, remove SetParams

sebastianst:


tee/prototype/enclave.go
```diff
@@ -143,10 +146,10 @@ func (e *Enclave) Init() (tee common.Address, _ []byte, err error) {
 	return e.account.Address, nil, nil
 }
 
-// SetParams should be called by the operator after they deployed the
+// setParams should be called by the operator after they deployed the
```
ndzik: Maybe adjust the comment. It's not a public method anymore.

PR #33, opened 2020-10-29
[cmd/operator] Add flag log-level

matthiasgeihs:


cmd/operator/main.go
```diff
@@ -10,13 +10,20 @@ import (
...
 	flag.Parse()
 
 	cfg, err := operator.LoadConfig(*configFilePath)
 	operator.AssertNoError(err)
 	log.Println("Config loaded")
 
-	_operator, _ := operator.Setup(cfg)
+	lvl, err := log.ParseLevel(*logLevel)
+	if err != nil {
+		log.Fatalf("parsing log level: %w", err)
```
sebastianst:
```suggestion
		log.Fatalf("parsing log level: %v", err)
```
`%w` only works with `fmt.Errorf` error wrapping when passing errors up the stack.

sebastianst (changes requested):
looking good, one minor issue

cmd/operator/main.go
matthiasgeihs: got it, done

sebastianst (approved):

PR #34, opened 2020-10-29
tee/prototype Add infolog for txCounter

ndzik:

### DIFF
--- a/tee/prototype/epochproc.go
+++ b/tee/prototype/epochproc.go
@@ -85,10 +85,12 @@ func (e *Enclave) txRoutine(
 		log = log.WithField("epoch", txEpoch.Number)
 	}
 
+	txCounter := 0
 	stagedTxs := makeTxCache()
 	for {
 		select {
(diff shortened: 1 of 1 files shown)


tee/prototype/epochproc.go
```diff
@@ -85,10 +85,12 @@ func (e *Enclave) txRoutine(
 		log = log.WithField("epoch", txEpoch.Number)
 	}
 
+	txCounter := 0
 	stagedTxs := makeTxCache()
 	for {
 		select {
 		case exiters := <-exits:
+			log.Infof("received %d TXs", txCounter)
```
sebastianst:
```suggestion
			log.Infof("txRoutine: processed %d TXs", txCounter)
```
 so we know where this log comes from.

ndzik: I even deleted `txRoutine:` before I submitted cause I thought it was only necessary for `trace`.
:D

sebastianst (approved):

PR #37, opened 2020-10-29
tee/epochproch Ease error handling

ndzik:

### DIFF
--- a/tee/prototype/epochproc.go
+++ b/tee/prototype/epochproc.go
@@ -112,17 +112,17 @@ func (e *Enclave) txRoutine(
 
 		case tx := <-txs:
 			txCounter++
-			log.WithFields(logrus.Fields{
+			log := log.WithFields(logrus.Fields{
 				"sender":    tx.Sender.String(),
 				"recipient": tx.Recipient.String(),
 				"nonce":     tx.Nonce,
 				"amount":    tx.Amount,
-			}).Debug("txRoutine: TX received, applying...")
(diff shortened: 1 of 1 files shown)


tee/prototype/epochproc.go
```diff
@@ -101,7 +101,7 @@ func (e *Enclave) txRoutine(
 
 			bps, err := e.generateBalanceProofs(txEpoch)
 			if err != nil {
-				return fmt.Errorf("generating balance proofs: %w", err)
+				log.Errorf("generating balance proofs: %w", err)
```
ndzik:
```suggestion
				log.Errorf("txRoutine: generating balance proofs: %v", err)
```

sebastianst: if this errors, it's really broken, so better leave as is. (note that you shouldn't use `%w` in logging)

tee/prototype/epochproc.go
```diff
@@ -122,7 +122,6 @@ func (e *Enclave) txRoutine(
 			err := e.applyEpochTx(txEpoch, tx)
 			if err != nil {
 				log.Errorf("txRoutine: Error applying tx: %v", err)
```
sebastianst, line 125: do `log := log.WithFields` in line 115 then we have nice context information here ;)

sebastianst (changes requested):
1 nicety ;)

sebastianst (approved):

PR #38, opened 2020-10-30
Demo

matthiasgeihs:
Prepare demo:
- add accountant for watching user and contract on-chain balances
- add demo readme and config
- extend bench command


demo/config.json
```diff
@@ -0,0 +1,10 @@
+{
+    "EthereumNodeURL": "ws://127.0.0.1:8545",
+	"Mnemonic": "pistol kiwi shrug future ozone ostrich match remove crucial oblige cream critic",
+	"EnclaveDerivationPath": "m/44'/60'/0'/0/0",
+	"OperatorDerivationPath": "m/44'/60'/0'/0/0",
+	"PhaseDuration": 3,
+	"ResponseDuration": 1,
```
sebastianst: set to 0 so exits always work in demo

demo/config.json
```diff
@@ -0,0 +1,10 @@
+{
+    "EthereumNodeURL": "ws://127.0.0.1:8545",
+	"Mnemonic": "pistol kiwi shrug future ozone ostrich match remove crucial oblige cream critic",
+	"EnclaveDerivationPath": "m/44'/60'/0'/0/0",
```
sebastianst, line 4: why not use /3 or higher for the enclave instead of a shared key with the op?

sebastianst: ahh I think we hardcode the enclave wallet derivation path to /0 right now, right?

client/client.go
```diff
@@ -231,22 +231,35 @@ func (c *Client) createTx(receiver common.Address, amount *big.Int) (tee.Transac
...
+		amount, ok = new(big.Int).SetString(args[2], 10)
+		if !ok {
+			status <- &CmdStatus{Err: fmt.Errorf("Could not parse <amount> as big.Int: %s", args[2])}
+			return
+		}
+	}
 	status <- &CmdStatus{Msg: fmt.Sprintf("Sending %d payments", n)}
 	result, err := Benchmark(n, func() error {
-		tx, err := c.createTx(c.Address(), big.NewInt(1)) // 1 WEI
+		tx, err := c.createTx(a, amount) // 1 WEI
```
sebastianst: fix comment, it's not `1 WEI` any more. Could we also divide the total amount by `n`? That would be better to enter, so I can benchmark "send 10 eth in 1000 microtransactions" with "bench 1000 0xRecv 10".

sebastianst: You can use the converters in `eth/units.go` so that would be something like
`singleTxAmount := new(big.Int).Div(eth.EthToWeiInt(amount), big.NewInt(n))`

demo/config.json
matthiasgeihs: done

sebastianst (approved):

PR #46, opened 2020-11-04
GitHub CI

### LINKED ISSUE #45: Setup CI
Taking some standard CI config from some of our public Go projects, including
* tests
* vetting

ggwpez:


operator/init_test.go
```diff
@@ -3,6 +3,6 @@ package operator
 import "github.com/sirupsen/logrus"
 
 func init() {
-	logrus.SetLevel(logrus.InfoLevel)
+	logrus.SetLevel(logrus.TraceLevel)
```
sebastianst: I wouldn't change that in master.

operator/config.go
```diff
@@ -33,3 +35,24 @@ func LoadConfig(path string) (*Config, error) {
 
 	return &config, nil
 }
+
+// envGanacheCmdVar is the environment variable to set the command
+// for spawning a `ganache-cli`. It is set "ERDSTALL_GANACHE_CMD".
+// See `ganacheCommand()`.
+const envGanacheCmdVar = "ERDSTALL_GANACHE_CMD"
```
sebastianst: The `Var` is superfluous.
```suggestion
const envGanacheCmd = "ERDSTALL_GANACHE_CMD"
```

operator/operator_test.go
```diff
@@ -122,18 +122,23 @@ func initEnvironment(t *testing.T) *environment {
 	var cmd *exec.Cmd
 	runGanache := true
 	if runGanache {
```
sebastianst: Please boyscout-fix this. Either just remove the bool and execute the indented block directly or pass `runGanache bool` as a boolean flag to `initEnvironment`. Not sure if there's a use case where we don't want to start it though. @matthiasgeihs what was the intention here?

operator/operator_test.go
```diff
@@ -178,7 +183,9 @@ func (e *environment) WaitPhase() {
 
 func (e *environment) Shutdown() {
 	if e.cmd != nil {
-		e.cmd.Process.Kill()
+		if err := e.cmd.Process.Kill(); err != nil {
+			log.Warn("Could not kill process:", err)
+		}
```
sebastianst, line 216: you could use `e.t` to assert that this doesn't fail.

operator/test/user.go
```diff
@@ -63,6 +63,7 @@ func CreateUser(
 ) *User {
 	ethClient, err := eth.CreateEthereumClient(ethURL, wallet, account)
 	if err != nil {
+		panic(err)
 		t.Fatal("creating ethereum wallet and client:", err)
```
sebastianst: Now the `Fatal` call is not reached. `Fatal` would crash the application anyways, why the panic?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
...
+        go-version: ^1.15
+
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Dependencies
+      run: |
+        go get -v -t -d ./...
+        docker pull trufflesuite/ganache-cli
+
```
sebastianst, line 34: We should also add `golangci-lint` here.

sebastianst: With a configuration similar to our go-perun one.

operator/operator_test.go
```diff
@@ -138,7 +138,7 @@ func initEnvironment(t *testing.T) *environment {
 		if err != nil {
 			panic(err)
 		}
-		time.Sleep(5 * time.Second)
+		time.Sleep(10 * time.Second)
```
sebastianst: Instead of a blind sleep, we could run a loop here that tries to connect to the ganache in, say .2 sec steps, for 10 sec. Would speed up the test in most cases. Or maybe instead of a loop, just one request with a 10 sec timeout, not sure what works best.

You could just do a generic query like `curl -X POST --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":99999}'`, see https://eth.wiki/json-rpc/API#eth_blocknumber or `eth_protocolVersion`, but not using curl but Go's internal http requests.

client/client.go
```diff
@@ -369,24 +364,6 @@ func (c *Client) CmdDeposit(status chan *CmdStatus, args ...string) {
 	}
 	// TODO challenge
 	status <- &CmdStatus{Err: errors.New("TODO challenge")}
-	return
-}
-
-func errorReason(ctx context.Context, b *pethchannel.ContractBackend, tx *types.Transaction, blockNum *big.Int, acc accounts.Account) (string, error) {
```
sebastianst, line 375: do you already use our `ContractBackend.WaitMined` function, which tries to extract revert reasons?

operator/operator_test.go
```diff
@@ -183,9 +183,7 @@ func (e *environment) WaitPhase() {
 
 func (e *environment) Shutdown() {
 	if e.cmd != nil {
-		if err := e.cmd.Process.Kill(); err != nil {
-			log.Warn("Could not kill process:", err)
-		}
+		e.cmd.Process.Kill() // nolint: errcheck
```
sebastianst: What's the linter's problem here? Wouldn't it be better to log the error?

sebastianst (changes requested):

operator/operator_test.go
ggwpez: That would actually have been much easier than to inject the spawn command via an `env` variable...  
I will remove it for now since it works and is easier to use. When we want it in the future we can still add it.   

Let's keep this open until @matthiasgeihs replies.

operator/operator_test.go
ggwpez: Good idea, i will look into it once i fixed all the other things.

client/client.go
ggwpez: This function was unused, therefore i deleted it. The client+operator currently use `abi/bind.WaitMined` but should probably switch to `ContractBackend.ConfirmTransaction`.

operator/operator_test.go
ggwpez: It collided with a change from Matthias, i will opt for the log.

operator/operator_test.go
ggwpez: This will be executed in the test cleanup, so i dont think we should/need to let it fail here.

operator/operator_test.go
sebastianst: I think it's still best to pass the ganach cmd as an env variable because that's easiest to configure in CI configs (like in your github one)

client/client.go
sebastianst: and we should probably just rename it to `WaitMined` in go-perun :joy:

sebastianst:
You can delete the Travis pipeline.

operator/operator_test.go
matthiasgeihs: this was for local testing reasons when i wanted to run ganache separately. i'm fine with removing that option.

ggwpez: Ok good

.github/workflows/ci.yml
ggwpez: I used the default config now, should be good.

operator/operator_test.go
ggwpez: I added a retry loop for the dial.

cmd/client/main.go
```diff
@@ -5,6 +5,7 @@ package main
 import (
 	perunchannel "perun.network/go-perun/backend/ethereum/channel"
 	perunhd "perun.network/go-perun/backend/ethereum/wallet/hd"
+	"perun.network/go-perun/log"
```
sebastianst: we use logrus, not perun log

operator/operator_test.go
```diff
@@ -119,21 +119,21 @@ func initEnvironment(t *testing.T) *environment {
...
+	args = append(args,
+		"--accounts=10",
+		"--defaultBalanceEther=100",
+		fmt.Sprintf("--mnemonic=\"%s\"", cfg.Mnemonic),
+		fmt.Sprintf("--blockTime=%d", blockTime),
+	)
+	cmd := exec.Command(
+		prog,
+		args...,
+	)
```
sebastianst: merge to one line

client/client.go
sebastianst: -> #54

sebastianst (changes requested):

operator/operator_test.go
```diff
@@ -112,7 +112,7 @@ type environment struct {
 	enclaveParameters tee.Parameters
 }
 
-const blockTime = 1 // block mining interval in seconds
+const blockTime = 2 // block mining interval in seconds
```
sebastianst: Can we change the test so that we subscribe to new heads to synchronize with the ganache's blocks? This way we can reuse the same test for testnets.

And I think that it makes sense to have the ganache use a block time of 1 sec so that tests run faster.

ggwpez: Yea we have a `SubscribeToBlocks` function in the ethereum client which should work.

eth/client.go
```diff
@@ -105,12 +106,21 @@ func (cl *Client) NewTransactor(ctx context.Context) (*bind.TransactOpts, error)
...
-	if err != nil {
-		return nil, fmt.Errorf("dialing ethereum: %w", err)
+func CreateEthereumClient(ctx context.Context, url string, wallet accounts.Wallet, a accounts.Account) (*Client, error) {
+	for {
+		ethClient, err := ethclient.DialContext(ctx, url)
+
+		if err != nil {
+			select {
+			case <-ctx.Done():
+				return nil, fmt.Errorf("dialing ethereum node: %v", err)
```
sebastianst:
```suggestion
				return nil, fmt.Errorf("dialing ethereum node: %w", err)
```

operator/operator_test.go
```diff
@@ -164,11 +164,17 @@ func initEnvironment(t *testing.T) *environment {
 }
 
 func (e *environment) WaitPhase() {
-	waitBlock := func() {
-		time.Sleep(blockTime * time.Second)
+	sub, err := e.operator.ethClient.SubscribeToBlocks()
```
sebastianst: This blocks subscription is unnecessarily complex because it retrieves _all_ receipts of the whole block. Please just make it work with a simple header subscription, the only information you need is the block number.

operator/operator_test.go
```diff
@@ -164,11 +164,17 @@ func initEnvironment(t *testing.T) *environment {
...
+	if err != nil {
+		e.T.Fatal("subscribing to blocks: ", err)
 	}
-	for i := uint64(0); i < e.enclaveParameters.PhaseDuration; i++ {
-		waitBlock()
+	defer sub.Unsubscribe()
+
+	for block := range sub.Blocks() {
+		// - 1 to make sure that we are in the new phase.
+		if e.enclaveParameters.IsLastPhaseBlock(block.NumberU64() - 1) {
```
sebastianst: How is `WaitPhase` used? It's confusing to see this `-1` because now you're waiting until the first block of a new phase has been mined, not when a phase starts.

sebastianst (changes requested):

operator/operator_test.go
ggwpez: I tried it before with a header subscription, but it did not work. I can try it again if you think it is worth it.

operator/operator_test.go
ggwpez: It did not work without the `-1`. Maybe @matthiasgeihs can tell us what it was actually meant for.

operator/operator_test.go
ggwpez: Changed it back to 1 sec.

operator/operator_test.go
ggwpez: Ok it seems to work now, i probably did not have the `-1` there while trying it before.

operator/operator_test.go
sebastianst: I implemented `IsLastPhaseBlock` and it does
```go
// IsLastPhaseBlock tells whether this block is the last block of a phase.
func (p Parameters) IsLastPhaseBlock(blockNum uint64) bool {
	return ((blockNum - p.InitBlock) % p.PhaseDuration) == p.PhaseDuration-1
}
```
so from my understanding, it is `true` for the last block of any phase. I.e., if the last block is _mined_ then the phase is done.

operator/operator_test.go
```diff
@@ -164,11 +166,25 @@ func initEnvironment(t *testing.T) *environment {
 }
 
 func (e *environment) WaitPhase() {
-	waitBlock := func() {
-		time.Sleep(blockTime * time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), blockTime*time.Duration(e.cfg.PhaseDuration+1))
```
sebastianst, line 169: PhaseDuration is in number of blocks _not_ a time. This here would be a few milli seconds, so always be a done context the moment you use it.

operator/operator_test.go
```diff
@@ -178,7 +182,36 @@ func (e *environment) WaitPhase() {
...
+		e.T.Fatal("subscribing to header: ", err)
+	}
+	defer sub.Unsubscribe()
+
+	for {
+		select {
+		case head := <-heads:
+			n--
+			if n == 0 {
+				log.Warn("Phase ended in block #", head.Number.Uint64())
```
sebastianst: I wouldn't put a `Warn` log here. Also, `WaitBlock` doesn't know why it is called, so it cannot say "Phase ended".

operator/operator_test.go
```diff
@@ -178,7 +182,36 @@ func (e *environment) WaitPhase() {
 	for {
 		select {
 		case head := <-heads:
-			if e.enclaveParameters.IsLastPhaseBlock(head.Number.Uint64() - 1) {
+			if e.enclaveParameters.IsLastPhaseBlock(head.Number.Uint64()) {
+				log.Warn("Phase ended in block #", head.Number.Uint64())
```
sebastianst: debug log

sebastianst (changes requested):

operator/operator_test.go
ggwpez: The stupid `Warn` calls are only my way of Debug logging since i do not need to change the loglevel to see them :laughing:  
Will surely remove them.

operator/operator.go
```diff
@@ -95,6 +95,7 @@ func Setup(cfg *Config) *Operator {
 		PhaseDuration:    cfg.PhaseDuration,
 		ResponseDuration: cfg.ResponseDuration,
 		PowDepth:         cfg.PowDepth,
+		InitBlock:        1,
```
sebastianst: You don't need that, `DeployContracts` should set the `InitBlock`. Also, you cannot know the correct init block number here.

operator/init_test.go
ggwpez: I just saw that the `tee/prototype/init_test.go` is set to `TraceLevel`, we could also change it there.

sebastianst (changes requested):

operator/operator.go
ggwpez: I will remove it.

sebastianst (approved):

PR #47, opened 2020-11-04
[prototype] Fix txCache append

ndzik:


ndzik:
~~This is only the fix for the `txCache`. The test will be included when also testing for `exits/withdrawals` in another PR.~~

Now also includes regression tests

Issue #42, opened 2020-11-04
Implement challenge/response in client/operator


sebastianst:
The operator should create a subscription to challenges and immediately answer to them.

The client should have a new `challenge` cmd which sends a challenge on-chain. Note that opening a challenge automatically exits the user from the system once the operator sends the proof to the contract (answering to challenges happens via the same `exit` call). So the `challenge` cmd should not only call the `challenge` function on the contract, but then also subscribe to a corresponding `Exiting` event. If the event arrives, wait for the end of the epoch and then call withdraw. So this is the same logic as for normal exits, so code duplication can be avoided.

I think for now we don't need to let the client automatically send challenges if they don't receive a proof in time, as this is just a demo. We can do it in a later issue if we receive interest in this.

PR #48, opened 2020-11-05
Operator challenge response

matthiasgeihs:
Addressing the operator part of #42

### DIFF
--- a/operator/operator.go
+++ b/operator/operator.go
@@ -7,7 +7,6 @@ import (
 	"net"
 	"net/http"
 	"net/rpc"
-	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -130,6 +129,8 @@ func (operator *Operator) Serve(port int) error {
 
 	// Handle on-chain challenges
 	//TODO
+	// errGo("Op.Challenges", operator.handleChallenges)
+	// log.Info("Operator.Serve: Challenge handling started")
 
 	// Handle RPC
 	errGo("Op.RPCServe", func() error { return operator.handleRPC(port) })
@@ -199,67 +200,6 @@ func (operator *Operator) handleRPC(port int) error {
 	return http.Serve(l, nil)
 }
 
-type depositProofs struct {
-	mu      sync.RWMutex
-	entries map[common.Address]*tee.DepositProof
-}
-
-func newDepositProofs() *depositProofs {
-	return &depositProofs{entries: make(map[common.Address]*tee.DepositProof)}
-}
-
-// Get gets the deposit proof for the given user, threadsafe.
-func (dps *depositProofs) Get(user common.Address) (*tee.DepositProof, bool) {
-	dps.mu.RLock()
-	defer dps.mu.RUnlock()
-
-	dp, ok := dps.entries[user]
-
-	return dp, ok
-}
-
-// AddAll adds the given deposit proofs, threadsafe.
-func (dps *depositProofs) AddAll(in []*tee.DepositProof) {
-	dps.mu.Lock()
-	defer dps.mu.Unlock()
-
-	for _, dp := range in {
-		dps.entries[dp.Balance.Account] = dp
-	}
-}
-
-type balanceProofs struct {
-	mu      sync.RWMutex
-	entries map[common.Address]*tee.BalanceProof
-}
-
-func newBalanceProofs() *balanceProofs {
-	return &balanceProofs{entries: make(map[common.Address]*tee.BalanceProof)}
-}
-
-// Get gets the balance proof for the given user, threadsafe.
-func (bps *balanceProofs) Get(user common.Address) (*tee.BalanceProof, bool) {
-	bps.mu.RLock()
-	defer bps.mu.RUnlock()
-
-	bp, ok := bps.entries[user]
-
-	return bp, ok
-}
-
-// AddAll adds the given balance proofs, threadsafe.
-func (bps *balanceProofs) AddAll(in []*tee.BalanceProof) {
-	bps.mu.Lock()
-	defer bps.mu.Unlock()
-
-	for _, bp := range in {
-		bps.entries[bp.Balance.Account] = bp
-	}
-}
-
-const gasLimit = 2000000
-const defaultContextTimeout = 10 * time.Second
-
 // AssertNoError logs the error and exits if the error is not nil.
 func AssertNoError(err error) {
 	if err != nil {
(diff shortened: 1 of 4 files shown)


operator/operator.go
```diff
@@ -115,150 +117,172 @@ func (operator *Operator) Serve(port int) error {
...
+					return
+				}
+
+				if r.Status != types.ReceiptStatusSuccessful {
+					//TODO: extract and log failure reason
+					log.Errorf("Operator.handleChallenges: Failed to complete response transaction for challenge %v", c)
+					return
+				}
+
+				log.Infof("Operator.handleChallenges: Resolved dispute for challenge %v", c)
```
ggwpez: I just had a quick look at it after rebasing my client onto it, and it seems this does not format too well :laughing:.  
It outputs:  
```pre
INFO[0069] Operator.handleChallenges: Resolved dispute for challenge &{2 [18 166 197 36 199 32 117 157 39 82 54 165 114 81 52 235 28 71 73 221] {[79 184 99 122 253 40 73 42 50 9 1 117 86 233 93 194 248 8 109 219] [[159 113 104 110 158 46 237 10 10 153 52 11 28 59 35 3 105 242 85 177 212 82 19 12 234 213 79 131 8 101 77 253] [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 2] [0 0 0 0 0 0 0 0 0 0 0 0 18 166 197 36 199 32 117 157 39 82 54 165 114 81 52 235 28 71 73 221]] [] 13 [195 37 53 68 40 35 185 152 139 79 218 227 201 66 145 88 129 56 47 74 202 233 131 209 9 207 60 114 112 198 47 111] 0 [82 70 186 155 7 187 173 137 50 186 127 193 185 233 248 18 56 45 170 36 190 75 227 75 247 197 80 132 83 249 54 44] 0 false}} 
```

operator/operator.go
```diff
@@ -160,6 +161,75 @@ func (operator *Operator) handleBlocks() error {
...
+
+		case c := <-challenges:
+			log.Warnf("Operator.handleChallenges: Incoming challenge for account %v and epoch %v", c.Account.String(), c.Epoch)
+
+			tr, err := func() (*bind.TransactOpts, error) {
+				ctx, cancel := createDefaultContext()
+				defer cancel()
+
+				return operator.ethClient.NewTransactor(ctx, big.NewInt(0), eth.DefaultGasLimit, operator.ethClient.Account())
+			}()
```
sebastianst: Don't you return a transactor here whose context is already closed when it's used?

operator/operator.go
```diff
@@ -160,6 +161,75 @@ func (operator *Operator) handleBlocks() error {
...
+		return fmt.Errorf("creating challenge subcription: %w", err)
+	}
+	defer sub.Unsubscribe()
+
+	for {
+		select {
+		case err := <-sub.Err():
+			return fmt.Errorf("subscription error: %w", err)
+
+		case c := <-challenges:
```
sebastianst: You could move the whole body of this `case` into its own method `handleChallenge(c *bindings.ErdstallChallenged) error` on the `*Operator` which would help readability because you save 2 levels of indentation. Then you can also replace all `log.Errorf(...); continue` statements with just `return fmt.Errorf(...)` and do the `log.Errorf(...); continue` here where you call the single challenge handler.

operator/operator.go
```diff
@@ -160,6 +161,75 @@ func (operator *Operator) handleBlocks() error {
...
+			}
+
+			tx, err := operator.contract.Exit(tr, balance, balanceProof.Sig)
+			if err != nil {
+				log.Errorf("Operator.handleChallenges: Failed to send response to challenge %v: %v", c, err)
+				continue
+			}
+
+			// Track challenge response transaction status
+			//TODO: do not create a new go routine for every transaction, run a dedicated go routine that tracks all?
```
sebastianst: I think it's fine how you do it for now. There shouldn't be that many challenges happening at the same time that this becomes an issue at this stage.

sebastianst (changes requested):

operator/operator.go
matthiasgeihs: Added formatted printing.

operator/operator.go
matthiasgeihs: I checked the implementation of `NewTransactor`. It only uses the context during creation of the transactor (to retrieve nonce and gas price). The created transactor doesn't rely on the context afterwards.

sebastianst: I see, also checked it. Then we have another problem: our`TransactOpts` never have any context set :joy: so there's no timeout on any blockchain requests that we do with those transactors.

sebastianst: So best solution now is to move the single challenge handling into its own method, as proposed above, and then use the same context for creating the transactor as well as setting it as the `Ctx` field, so that below `Exit` call has the context set. Putting it into its own method ensures that the `defer cancel()` calls don't accumulate in the loop and would only be called all at once when this loop stops (i.e., at the end of the routine/program).

operator/operator.go
matthiasgeihs: done

operator/operator.go
matthiasgeihs: Agreed, we don't have to take care of it now. However, I would like to keep note of that so that we do not forget about it later. I would therefore keep the comment, or, alternatively, creating an issue for hat.

operator/operator.go
matthiasgeihs: ok, done

operator/operator.go
matthiasgeihs: Created #50

sebastianst: I prefer creating an issue so we don't have rotting TODO comments.

operator/operator.go
```diff
@@ -160,6 +162,78 @@ func (operator *Operator) handleBlocks() error {
...
+		}
+	}
+}
+
+func (operator *Operator) handleChallengedEvent(c challengedEvent) error {
+	ctx, cancel := createDefaultContext()
+	defer cancel()
+
+	tr, err := operator.ethClient.NewTransactor(ctx, big.NewInt(0), eth.DefaultGasLimit, operator.ethClient.Account())
+	tr.Context = ctx
```
sebastianst: move below `if err` check or this might panic due to a `nil` `tr`.

sebastianst (changes requested):

operator/operator.go
matthiasgeihs: done

sebastianst (approved):

PR #49, opened 2020-11-05
Enclave exit withdraw test

### LINKED ISSUE #40: Enclave: Test exits and withdrawals
For this, the client/test Client needs to be extended to also support sending exit and withdrawal transactions.

Can then also use this in #39

ndzik:
Closes #40

### DIFF
--- a/client/test/client.go
+++ b/client/test/client.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/perun-network/erdstall/contracts/bindings"
@@ -64,25 +65,32 @@ func (c *Client) UpdateLastBlockNum() {
 }
 
 // Deposit deposits amount to the Erdstall contract and waits until the deposit
-// tx is mined. The minedBlockNum is updated automatically.
+// TX is mined. The minedBlockNum is updated automatically.
 func (c *Client) Deposit(ctx context.Context, amount *big.Int) error {
-	tr, err := c.ethClient.NewTransactor(ctx, amount, eth.DefaultGasLimit, c.ethClient.Account())
-	if err != nil {
-		return fmt.Errorf("creating transactor: %w", err)
-	}
+	return c.call(ctx, func(tr *bind.TransactOpts) (*types.Transaction, error) {
+		tr.Value = amount
+		return c.contract.Deposit(tr)
+	})
+}
 
-	tx, err := c.contract.Deposit(tr)
-	if err != nil {
-		return fmt.Errorf("calling deposit: %w", err)
-	}
+// Exit exits the Erdstall contract and waits until the exit TX is mined. The
+// minedBlockNum is updated automatically.
+func (c *Client) Exit(ctx context.Context, bal *tee.BalanceProof) error {
+	return c.call(ctx, func(tr *bind.TransactOpts) (*types.Transaction, error) {
+		return c.contract.Exit(tr, bindings.ErdstallBalance{
+			Epoch:   bal.Balance.Epoch,
+			Account: bal.Balance.Account,
+			Value:   bal.Balance.Value,
+		}, bal.Sig)
+	})
+}
 
-	rec, err := bind.WaitMined(ctx, c.ethClient, tx)
-	if err != nil {
-		return fmt.Errorf("waiting for deposit block: %w", err)
-	}
-	c.SetMinedBlockNum(uint64(rec.BlockNumber.Int64()))
-
-	return nil
+// Withdraw withdraws the balance with the given `tee.BalanceProof` and waits
+// until the withdraw TX is mined. The minedBlockNum is updated automatically.
+func (c *Client) Withdraw(ctx context.Context, bal *tee.BalanceProof) error {
+	return c.call(ctx, func(tr *bind.TransactOpts) (*types.Transaction, error) {
+		return c.contract.Withdraw(tr, bal.Balance.Epoch)
+	})
 }
 
 func (c *Client) Send(recipient common.Address, amount *big.Int) error {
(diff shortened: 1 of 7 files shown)


tee/prototype/enclave_internal_test.go
```diff
@@ -126,6 +134,63 @@ func TestEnclave(t *testing.T) {
 		requiree.NoError(err)
 	}
 
+	doWith := func(bp *tee.BalanceProof, aliceDo, bobDo cltest.ClientAction) {
+		switch bp.Balance.Account.String() {
+		case aliceAd.Address.String():
+			requiree.NoError(aliceDo(ctx, bp))
+		case bobAd.Address.String():
+			requiree.NoError(bobDo(ctx, bp))
```
sebastianst: Can't you switch on a `[32]byte` directly?

tee/prototype/enclave_internal_test.go
```diff
@@ -126,6 +134,63 @@ func TestEnclave(t *testing.T) {
 		requiree.NoError(err)
 	}
 
+	doWith := func(bp *tee.BalanceProof, aliceDo, bobDo cltest.ClientAction) {
```
sebastianst: You define `ClientAction` in the prior commit but only use it here and it somewhat stands lonely in the `client/test` package. Maybe just define it privately here in this test package?

tee/prototype/enclave_internal_test.go
```diff
@@ -126,6 +134,63 @@ func TestEnclave(t *testing.T) {
...
+
+	seal("exitPhase", 1)
+
+	t.Log("Sending two withdrawal TXs.")
+	for _, bp := range bps {
+		doWith(bp, alice.Withdraw, bob.Withdraw)
+	}
+
+	aliceNewBal, _ := setup.SimBackend.BalanceAt(ctx, aliceAd.Address, nil)
+	bobNewBal, _ := setup.SimBackend.BalanceAt(ctx, bobAd.Address, nil)
```
sebastianst: Why do you ignore the errors instead of `requiree.NoError`?

tee/prototype/enclave_internal_test.go
```diff
@@ -126,6 +134,63 @@ func TestEnclave(t *testing.T) {
...
+	requiree.NoError(checkBals(aliceInitBal, aliceNewBal, bobInitBal, bobNewBal, 3))
+
 	sub.Unsubscribe()
 	ct.Wait("operator")
 }
+
+// checkBals checks whether the new balance of Alice has increased by `difference`
+// and conversely whether the new balance of Bob has decreased.
```
sebastianst: mention that `difference` is denominated in eth (or pass it as a `big.Int` in wei) because right now it's a bit confusing that you pass 4 numbers as wei `big.Int`s and one number as `int64` ether.

client/test/client.go
```diff
@@ -9,23 +9,28 @@ import (
...
+		tr        Transactor
 
-	Nonce         uint64
-	minedBlockNum uint64
-}
+		Nonce         uint64
+		minedBlockNum uint64
+	}
+
+	ClientAction = func(ctx context.Context, bal *tee.BalanceProof) error
```
sebastianst: See other comment, I'd change this to a test-local private

client/test/client.go
```diff
@@ -85,6 +90,55 @@ func (c *Client) Deposit(ctx context.Context, amount *big.Int) error {
...
+	tr, err := c.ethClient.NewTransactor(ctx, big.NewInt(0), eth.DefaultGasLimit, c.ethClient.Account())
+	if err != nil {
+		return fmt.Errorf("creating transactor: %w", err)
+	}
+
+	tx, err := c.contract.Exit(tr, bindings.ErdstallBalance{
+		Epoch:   bal.Balance.Epoch,
+		Account: bal.Balance.Account,
+		Value:   bal.Balance.Value,
+	}, bal.Sig)
```
sebastianst: You can avoid all code duplication between `Deposit`, `Exit` and `Withdraw` if you create a method `call(ctx context.Context, call func(ctx) (types.Transaction, error)` that contains the same body as this function but here just calls `tx, err := call(ctx)` and then use it to define `Deposit`, `Exit` and `Withdraw`.

This way we'd also fix a "bug" in `Deposit` that we don't check for a successful tx execution right now :)

sebastianst (changes requested):

tee/prototype/enclave_internal_test.go
ndzik: I thought this error would be caught anyway, when checking the balances. But I realize this is only by "chance" so better add those checks back.

client/test/client.go
```diff
@@ -66,23 +67,37 @@ func (c *Client) UpdateLastBlockNum() {
...
-	if err != nil {
-		return fmt.Errorf("creating transactor: %w", err)
-	}
-
-	tx, err := c.contract.Deposit(tr)
-	if err != nil {
-		return fmt.Errorf("calling deposit: %w", err)
-	}
+	return c.call(ctx, func(ctx context.Context) (*types.Transaction, error) {
+		tr, err := c.ethClient.NewTransactor(ctx, amount, eth.DefaultGasLimit, c.ethClient.Account())
```
ndzik: I opted for creating the transactor in each call (like you also suggested) but just in case this question might come up:

One could also dedup the transactor creation by letting the caller pass an extra `amount` to `Client.call(...)` but this broke one too many logical barriers imo, so I also settled for the current impl.

sebastianst: I just remembered that we don't set the contexts yet in `NewTransactor`, so everywhere where you create the `tr`, after the `err != nil` check, do `tr.Context = ctx`. It will be fixed in go-perun soon.

Even better yet, you could also create the method `NewTransactor` on the eth client and do it there, then we only need to change it in one place.

sebastianst: And go-perun will also soon remove the `amount` parameter, then it needs to be set manually using `tr.Amount = amount` anyways. So I think you can change the function passed to `call` to `func(context.Context, tr *bind.TransactOpts) (*types.Transaction, error)` and then set the amount on `tr` in `Deposit`.

eth/client.go
```diff
@@ -346,7 +359,7 @@ func (cl *Client) DeployContracts(params *tee.Parameters) error {
 	ctx, cancel := NewDefaultContext()
 	defer cancel()
 
-	tr, err := cl.NewTransactor(ctx, big.NewInt(0), DefaultGasLimit, cl.account)
+	tr, err := cl.ContractBackend.NewTransactor(ctx, big.NewInt(0), DefaultGasLimit, cl.account)
```
sebastianst: Why not use the new `NewTransactor` method? Would be better because the context is missing here again.

sebastianst (changes requested):
Please check all uses of `eth.Client.NewTransactor`, also in `client` and `operator`: you can remove setting the context manually since you now do it in the new client method.

client/test/client.go
```diff
@@ -112,3 +116,28 @@ func (c *Client) SignTx(tx *tee.Transaction) {
 		log.Panicf("Error signing tx: %v", err)
 	}
 }
+
+// call calls the given function, waits for the TX to be mined and updates the
+// clients last mined blocknumber.
+func (c *Client) call(ctx context.Context, call func(context.Context, *bind.TransactOpts) (*types.Transaction, error)) error {
```
sebastianst: Now that we have the context in `tr`, you can remove it from the `call func` parameters :see_no_evil: 
```suggestion
func (c *Client) call(ctx context.Context, call func(*bind.TransactOpts) (*types.Transaction, error)) error {
```

ndzik: Oh, right.

ndzik:
> Please check all uses of `eth.Client.NewTransactor`, also in `client` and `operator`: you can remove setting the context manually since you now do it in the new client method.

I was confused for a second, then realized I  did not fetch the `master` branch for some time. Hehe.

ndzik:
Squashed commits.

sebastianst (approved):

sebastianst (approved):

PR #53, opened 2020-11-10
tee/prototype: Return processing errors for ProcessTXs/Blocks

sebastianst:
Before, these functions just pushed the request into the processors'
queues and returned nil immediately. Now they block until all transactions
or blocks are processed and returns all errors, if any, that occurred
while processing.

### DIFF
--- a/client/test/client.go
+++ b/client/test/client.go
@@ -26,6 +26,7 @@ type Client struct {
 
 	Nonce         uint64
 	minedBlockNum uint64
+	balance       *big.Int // local balance tracking
 }
 
 // NewClient creates a testing Erdstall client.
@@ -42,12 +43,21 @@ func NewClient(params tee.Parameters, wallet tee.TextSigner, ethClient *eth.Clie
 		contract:  contract,
 		wallet:    wallet,
 		tr:        tr,
+		balance:   new(big.Int),
 	}, nil
 }
 
-func (c *Client) NextNonce() uint64 {
+func (c *Client) nextNonce() uint64 {
 	c.Nonce++
 	return c.Nonce
+}
+
+func (c *Client) Balance() *big.Int {
+	return new(big.Int).Set(c.balance)
+}
+
+func (c *Client) Address() common.Address {
+	return c.ethClient.Account().Address
 }
 
 func (c *Client) SetMinedBlockNum(n uint64) {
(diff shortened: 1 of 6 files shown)


tee/prototype/epochproc.go
```diff
@@ -120,9 +124,12 @@ func (e *Enclave) txRoutine(
 			})
 			log.Debug("txRoutine: TX received, applying...")
 			stagedTxs.cacheTx(tx)
-			err := e.applyEpochTx(txEpoch, tx)
-			if err != nil {
+			if err := e.applyEpochTx(txEpoch, tx); err != nil {
```
ndzik, line 127: I remember now, that our `applyEpochTx(...)` makes it way to easy to DOS our system. Unless we require the operator to pre-validate all TXs before pushing them into the Enclave, we shutdown way to strict in that regard.

I propose to have a Enclave.Global `error` type which signals a hard failure. Since we `wrap` our errors, we can easily check with `errors.Is(returnedErr, EnclaveAbortErr)` whether we need to stop the enclave or not. Would also allow to keep logging all errors and remove the workaround present in `applyEpochTx(...)` where we only allow mismatched epoch numbers to occur.

But this would probably be a new issue then.

sebastianst: The whole enclave design is not production ready as it is right now anyways ;) it's just still a demo, after all. I propose to discuss fundamental changes to how we deal with invalid txs in a new issue. This PR should only provide one thing, error/result receipts to the caller of `ProcessTXs`/`ProcessBlocks`.

(Pre-validation has the big disadvantage that we'd check all signatures twice on the same machine, since the op would share computational ressorces with the TEE. In a production setting this could be solved by a layered approach, where one part of the operator is running on a different machine, just receiving and validating txs. But this only becomes relevant in the far future, I guess ;) )

client/test/client.go
```diff
@@ -112,7 +133,90 @@ func (c *Client) Send(recipient common.Address, amount *big.Int) error {
...
+	txs = append(txs, &tx0)
+
+	signAppend := func(tx *tee.Transaction) {
+		c.SignTx(tx)
+		txs = append(txs, tx)
+	}
+
+	// 1. Invalid nonce
+	tx1 := valid
+	tx1.Nonce += 42
```
ndzik: I have the feeling it might be better to test the edge cases, where the `Nonce` (and everywhere else, where applicable) is exactly `1` more and/or less than expected.

Since off-by-one errors are more common and this would also catch places where an implementer mixed up `{<|>|>=|<=}`.

client/test/client.go
```diff
@@ -112,7 +133,90 @@ func (c *Client) Send(recipient common.Address, amount *big.Int) error {
...
+	signAppend(&tx1)
+
+	// 2. Invalid epoch
+	tx2 := valid
+	tx2.Epoch += 42
```
ndzik: See other comment for off-by-one errors.

eth/wallet.go
```diff
@@ -21,3 +24,7 @@ func NewHdWallet(rng io.Reader) *hdwallet.Wallet {
 	}
 	return hdw
 }
+
+func NewRandomAddress(rng *rand.Rand) common.Address {
```
ndzik: This function is also already implemented in `tee/test/transaction.go`. Would be nice to remove it from there if possible.

Would need to touch the `txCache` test then IIRC.

ndzik (changes requested):

client/test/client.go
sebastianst: good point, will do that

eth/wallet.go
sebastianst: Ah forgot about that one, will remove my implementation then.

client/test/client.go
sebastianst: This inspired me to also check for negative amounts and we didn't check that yet :joy: fixed now.

ndzik (approved):

PR #56, opened 2020-11-16
Dishonest operator

ggwpez:
The honest value in the config.json now decides whether the
operator should react to on-chain challenges with a response.

### DIFF
--- a/cmd/operator/config.json
+++ b/cmd/operator/config.json
@@ -6,5 +6,6 @@
 	"PhaseDuration": 3,
 	"ResponseDuration": 1,
 	"PowDepth": 0,
-	"Port": 8080
+	"Port": 8080,
+	"RespondChallenges": true
 }
--- a/config.json
+++ b/config.json
@@ -6,5 +6,6 @@
 	"PhaseDuration": 3,
 	"ResponseDuration": 0,
 	"PowDepth": 0,
-	"Port": 8080
+	"Port": 8080,
+	"RespondChallenges": false
 }
--- a/operator/config.go
+++ b/operator/config.go
@@ -16,6 +16,7 @@ type Config struct {
 	ResponseDuration       uint64
 	PowDepth               uint64
 	Port                   int
+	RespondChallenges      bool
 }
 
 // LoadConfig loads an operator configuration from the given file path.
--- a/operator/operator.go
+++ b/operator/operator.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	hdwallet "github.com/miguelmota/go-ethereum-hdwallet"
 	log "github.com/sirupsen/logrus"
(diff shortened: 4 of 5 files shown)


matthiasgeihs:
You could print a big fat warning when starting the operator in "dishonest" mode.

ggwpez:
> You could print a big fat warning when starting the operator in "dishonest" mode.

Yea good idea. I will add something.

cmd/operator/config.json
```diff
@@ -6,5 +6,6 @@
 	"PhaseDuration": 3,
 	"ResponseDuration": 1,
 	"PowDepth": 0,
-	"Port": 8080
-}
\ No newline at end of file
+	"Port": 8080,
+	"Honest": true
```
sebastianst: Rename to something more specific, like "RespondChallenges". Hones is quite vague and can mean many things, like not sending proofs etc. When we later add other protocol-deviation behavior, it helps naming the things more specifically.

config.json
```diff
@@ -6,5 +6,6 @@
 	"PhaseDuration": 3,
 	"ResponseDuration": 0,
 	"PowDepth": 0,
-	"Port": 8080
+	"Port": 8080,
+	"Honest": true
```
sebastianst: See other comment, rename to "RespondChallenges"

operator/config.go
```diff
@@ -16,6 +16,7 @@ type Config struct {
 	ResponseDuration       uint64
 	PowDepth               uint64
 	Port                   int
+	Honest                 bool
```
sebastianst: `RespondChallenges`

operator/operator.go
```diff
@@ -44,6 +45,7 @@ func New(
 	params tee.Parameters,
 	client *eth.Client,
 	contract common.Address,
```
sebastianst: contract is part of `params`, this parameter could be removed

operator/operator.go
```diff
@@ -31,6 +31,7 @@ type Operator struct {
 	*depositProofs
 	*balanceProofs
 	contract *bindings.Erdstall
+	honest   bool // Set iff the Operator behaves honest.
```
sebastianst: see other comment, rename to `respondChallenges`.

operator/operator.go
```diff
@@ -44,6 +45,7 @@ func New(
 	params tee.Parameters,
 	client *eth.Client,
 	contract common.Address,
+	honest bool,
```
sebastianst: see other comment, rename to `respondChallenges`.

operator/operator.go
```diff
@@ -187,6 +190,11 @@ func (operator *Operator) handleChallenges() error {
 }
 
 func (operator *Operator) handleChallengedEvent(c challengedEvent) error {
+	if !operator.honest {
+		log.Info("Operator.handleChallengedEvent: dishonestly ignoring challenge")
```
sebastianst:
```suggestion
		log.Warn("Operator.handleChallengedEvent: ignoring challenges, returning.")
```

operator/config.go
sebastianst: Could you also give it a cmd line flag or is it more work to combine cmd line flags and config in the current implementation?

sebastianst (changes requested):

operator/config.go
ggwpez: The current implementation does not have flags, so this would be the only option with a flag.  
I also dont know how well it integrates with the current config logic.

cmd/operator/config.json
ggwpez: Renamed it to `RespondChallenges`.

sebastianst (approved):

PR #59, opened 2020-11-17
Client Challenge + contract freezing

### LINKED ISSUE #42: Implement challenge/response in client/operator

ggwpez:


eth/client.go
```diff
@@ -277,18 +277,18 @@ func (cl *Client) SubscribeToBlocksStartingFrom(startBlockNumber *big.Int) (*Blo
 	return blockSub, nil
 }
 
-// SubscribeToDeposited writes past Deposited events and newly received ones
```
sebastianst, line 290: Please remove the superfluous `To` from all `Subscribe` functions.

eth/client.go
```diff
@@ -298,7 +298,45 @@ func (cl *Client) SubscribeToDeposited(ctx context.Context, contract *bindings.E
...
 	}
-	it, err := contract.FilterDeposited(fOpts, epochs, accs)
+	it, err := contract.FilterExiting(fOpts, epochs, accs)
+	defer it.Close()
+	if err != nil {
+		return err
+	}
+	for it.Next() {
+		sink <- it.Event
+	}
```
sebastianst: I think the way you do this now has a problem: there is a race between future events and past events from the filter iterator. If there is a future event coming in before the iterator is exhausted, the events are not in order. Same for the other sub.

eth/client.go
```diff
@@ -298,7 +298,45 @@ func (cl *Client) SubscribeToDeposited(ctx context.Context, contract *bindings.E
...
+	// Wait for error or ctx done.
+	select {
+	case err := <-sub.Err():
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
```
sebastianst: The ctx should only be used for setting up the sub, not for waiting. Also, the subscribe function now blocks until the sub is closed or ctx is done, which is uncommon. Shouldn't the subscribe function return a sub so that the caller can close the sub?

Same for other sub

client/client.go
```diff
@@ -421,86 +405,191 @@ func (c *Client) BalanceProofWatcher() {
...
+func (c *Client) FrozenWatcher() {
+	freeze := make(chan *bindings.ErdstallFrozen)
+	subErr := make(chan error)
+	ctx, cancel := context.WithCancel(c.Ctx())
+	defer cancel()
+	go func() {
+		subErr <- c.ethClient.SubscribeToFrozen(ctx, c.contract, nil, freeze)
+	}()
+
+	for {
```
sebastianst: I think you can remove the `for` loop, both signals can only happen at most once.

client/client.go
```diff
@@ -421,86 +405,191 @@ func (c *Client) BalanceProofWatcher() {
...
+						c.logOnChain(msg.Msg)
+					}
+				}
+			}()
+			_, err := c.sendTx("WithdrawFrozen", func(auth *bind.TransactOpts) (*types.Transaction, error) {
+				return c.contract.WithdrawFrozen(auth, bindings.ErdstallBalance{Epoch: epoch, Account: bal.Account, Value: bal.Value}, bal.Bal.Sig)
+			}, status)
+			if err != nil {
+				return
+			}
```
sebastianst: You don't wanna log/print the error?

client/client.go
```diff
@@ -421,86 +405,191 @@ func (c *Client) BalanceProofWatcher() {
...
+	go func() {
+		subErr <- c.ethClient.SubscribeToFrozen(ctx, c.contract, nil, freeze)
+	}()
+
+	for {
+		select {
+		case err := <-subErr:
+			c.logOnChain("Frozen event subscription: %v", err)
+			return
+		case event := <-freeze:
```
sebastianst: You may want to move the whole `case` body inside a `handleFrozen` method.

client/client.go
```diff
@@ -421,86 +405,191 @@ func (c *Client) BalanceProofWatcher() {
...
+
+	select {
+	case <-done: // ChallengeReponse phase is over, freeze.
+		c.log("🧊 Freezing Contract")
+		_, err := c.sendTx("Freeze", func(auth *bind.TransactOpts) (*types.Transaction, error) {
+			return c.contract.Freeze(auth)
+		}, status)
+		if err != nil {
+			return
+		}
```
sebastianst: log error?

client/client.go
```diff
@@ -421,86 +405,191 @@ func (c *Client) BalanceProofWatcher() {
...
+	case err := <-subError:
+		if err != nil {
+			c.logOnChain(err.Error())
```
sebastianst: If you log like this, it may not be clear that his is an error because you just log the error message as a string, not even prefixed with `Error: `. Applies everywhere where you just log an error. Maybe add a special logger method for that?

client/client.go
sebastianst: And add comment below that `FrozenWatcher` will handle withdrawal here

client/client.go
```diff
@@ -421,86 +405,191 @@ func (c *Client) BalanceProofWatcher() {
...
+		}
 		return
+	case <-exits: // Challenge was posted on-chain by OP.
+		c.logOnChain("Received on-chain challenge response")
```
sebastianst, line 539: After the op has answered the challenge, you still need to call `withdraw` with the correct balance proof, no?

tee/params.go
```diff
@@ -51,7 +51,7 @@ func (p Parameters) EpochStartBlock(epoch uint64) uint64 {
 	return p.InitBlock + epoch*p.PhaseDuration
 }
 
-func (p Parameters) DepositEndBlock(epoch uint64) uint64 {
+func (p Parameters) EpochEndBlock(epoch uint64) uint64 {
 	return p.EpochStartBlock(epoch) + p.PhaseDuration
 }
```
sebastianst: The naming of those last 3 methods is imprecise.
- `EpochStartBlock` should rather be `DepositStartBlock`
- `EpochEndBlock` should rather be `DepositDoneBlock` or `TxStartBlock`, depending on usage
- `TxEndBlock` should rather be `TxDoneBlock` or `ExitStartBlock`, depending on usage

sebastianst (changes requested):

client/client.go
ggwpez: <s>Yes this seems counterintuitive, i will let `sendTx` return a bool instead of an error, since it already logs the error itself. </s>
  
PS Added a `log` here, since there would be too much code duplication if `sendTx` logs it.

eth/client.go
ggwpez: Good catch, seems like this would indeed mess with the event order.

eth/client.go
ggwpez: You mean `<-ctx.Done()` should not be part of the select?  
We also do it like this in `go-perun` since i think that the go-ethereum subscription does not propagate up the correct context-cancelled error, but im not sure.

ggwpez: I can also change the subscribe functions to return a sub type, but i wrote them in parallel to yours, so they are designed differently :smile:

client/client.go
ggwpez: Yea smart

client/client.go
ggwpez: This probably did not happen yet so i did not see it :laughing:   
I added a `logError` method.

eth/client.go
sebastianst: I'm pretty sure I didn't write the other sub function. You can use `git blame` if you're interested ;)

Sure that we did it like this in `go-perun`? E.g. for `Adjudicator.Subscribe` we have our own subscription type which has an iterator API.

I think the API of these subscribe functions is a bit confusing, because you would assume the `SubscribeX` function to return after successfully setting up the sub. You could change it so that you create the channel internally and then return the channel and the `ethereum.Subscription`.

ggwpez: I know that it is a bad API, was just because of time pressure. I can change the Sub functions to all return immediately, will just take some time to refactor then.  
  
[Here](https://git.perun.network/perun/go-perun/-/blob/dev/backend/ethereum/channel/withdraw.go#L54) for example we also wait on the ctx that created the watch opts.

ggwpez: No idea who wrote the other sub functions :see_no_evil: thought it was you or Matthias.

client/client.go
ggwpez: Hm i did not even test this case yet ...

eth/client.go
ggwpez: So i will add `FrozenSubscription` and `ExitingSubscription` types.

eth/client.go
ggwpez: I simplified all event subscriptions to just use `startBlock=0` and removed the past querying logic.

client/client.go
ggwpez: Done

client/client.go
ggwpez: Added  
`// The FrozenWatcher will handle the withdrawal from hereon.`

eth/client.go
sebastianst: Did you check that this works? At least last year, when we were developing go-perun's ethereum backend, I remember that it unfortunately didn't work (yet) so we definitely have to test that this really works. Might even be provider dependent.

ggwpez: We could write a dedicated test for it, otherwise i did not encounter problems with it (yet).

eth/client.go
```diff
@@ -49,6 +50,16 @@ type (
...
+
+	ExitingSubscription struct {
+		sub    event.Subscription
+		events chan *bindings.ErdstallExiting
+	}
+
+	FrozenSubscription struct {
+		sub    event.Subscription
+		events chan *bindings.ErdstallFrozen
+	}
```
sebastianst, line 62: I'm so much looking forward to generics.

eth/client.go
```diff
@@ -287,42 +298,76 @@ func (cl *Client) SubscribeToBlocksStartingFrom(startBlockNumber *big.Int) (*Blo
...
+// Can be cancelled via Unsubscribe.
+func (cl *Client) SubscribeExiting(ctx context.Context, contract *bindings.Erdstall, epochs []uint64, accs []common.Address) (*ExitingSubscription, error) {
 	// sub to new events
 	wOpts, err := cl.NewWatchOpts(ctx)
 	if err != nil {
-		return err
+		return nil, err
 	}
-	sub, err := contract.WatchDeposited(wOpts, sink, epochs, accs)
+	start := uint64(0) // Begin the query at block 0.
```
sebastianst: The moment you're subscribing, aren't you actually only interested in all events since the start of the last phase? All events of any prior phase are irrelevant, no?

As discussed, not sure if this option even works for past blocks, or only for future blocks.

sebastianst: And even if you want to subscribe to _all_ events, you can still start at `InitBlock`, no need to start at 0, which would be quite unnecessary on a testnet or mainnet.

client/client.go
```diff
@@ -200,8 +200,8 @@ func (c *Client) CmdSend(status chan *CmdStatus, args ...string) {
 	}
 }
 
-func (c *Client) createTx(receiver common.Address, amount *big.Int) (tee.Transaction, error) {
-	block := atomic.LoadUint64(&c.currentBlock) + 1
+func (c *Client) createTransfer(receiver common.Address, amount *big.Int) (tee.Transaction, error) {
+	block := atomic.LoadUint64(&c.lastBlock) + 1
```
sebastianst, line 204: why the rename? you create a tx, not a transfer

client/client.go
```diff
@@ -386,6 +386,60 @@ func (c *Client) BalanceProofWatcher() {
...
+		c.logError("Frozen event subscription: %v", err)
+		return
+	case event := <-sub.Events():
+		c.handleFrozen(event)
+	}
+}
+
+func (c *Client) handleFrozen(event *bindings.ErdstallFrozen) {
+	epoch := event.Epoch
+	c.events <- &Event{Type: SET_OP_TRUST, OpTrust: UNTRUSTED}
```
sebastianst: you may also want to create a method for setting the op trust, like you did for the other `Event`s

client/client.go
```diff
@@ -386,6 +386,60 @@ func (c *Client) BalanceProofWatcher() {
...
+			if msg.Err != nil {
+				c.logOnChain(msg.Err.Error())
+			} else {
+				c.logOnChain(msg.Msg)
+			}
+		}
+	}()
+	_, err := c.sendTx("WithdrawFrozen", func(auth *bind.TransactOpts) (*types.Transaction, error) {
+		return c.contract.WithdrawFrozen(auth, bindings.ErdstallBalance{Epoch: epoch, Account: bal.Account, Value: bal.Value}, bal.Bal.Sig)
+	}, status)
```
sebastianst, line 435: Why do you even pass a `status` channel to `sendTx` instead of just using `c.logOnChain` in it? It looks funny that you start above this logger go routine even though you could just call the `c.logOnChain` methods directly where you write `CmdStatus`'s into the channel.

sebastianst: btw you're missing a newline above the def of method `sendTx`.

client/client.go
```diff
@@ -386,6 +386,60 @@ func (c *Client) BalanceProofWatcher() {
 	}
 }
 
+// FrozenWatcher listenes for Frozen events and calls WithdrawFrozen
+// when a balance proof is available.
```
sebastianst:
```suggestion
// FrozenWatcher listens for Frozen events and calls WithdrawFrozen
// if a balance proof is available.
```

config/client.go
```diff
@@ -14,7 +14,7 @@ type ClientConfig struct {
 
 func ParseClientConfig() (cfg ClientConfig) {
 	flag.StringVar(&cfg.ChainURL, "chain-url", "ws://127.0.0.1:8545", "'protocol://ip:port'")
-	flag.StringVar(&cfg.Mnemonic, "mnemonic", "pistol kiwi shrug future ozone ostrich match remove crucial oblige cream critic", "Wallet mnemonic.")
+	flag.StringVar(&cfg.Mnemonic, "mnemonic", "tag volcano eight thank tide danger coast health above argue embrace heavy", "Wallet mnemonic.")
```
sebastianst: why did you change the mnemonic from the readme tutorial?

sebastianst (changes requested):

eth/client.go
sebastianst: We should. Maybe in the cases that you tried, you always set up the sub early enough. But we cannot assume in general that this will be the case. Then we could also just omit the 0.

client/client.go
sebastianst: looking at the code further, it seems that you always pass the `status` channel down the stack?

eth/client.go
ggwpez: Yes this is horrible :rofl:

eth/client.go
ggwpez: The `InitBlock` should be used here indeed, but i dont know if we even need past events.  
I guess in most cases it does not matter, only in some corner cases when the client starts and the contract is already frozen or something.  
The `eth` client does not know the last block number, so it would have to calculate the first blocknumber that it should use for the query from the `epochs`  argument.

client/client.go
ggwpez: i thought `Transaction` sound more like a general on-chain transaction, whereas this creates a TEE balance transfer.

client/client.go
ggwpez: There are several output windows for this, the `status` is displayed in the lower right, which tracks the progress/status of a command whereas the `log` calls are not associated to a specific command.

config/client.go
ggwpez: Hm i changed it to the default mnemonic in the operator unaware that we also used it in the `README.md`. I will instead change the default mnemonic of the Operator.

ggwpez: I just understood that we have two `README` files with different mnemonics :see_no_evil: i will change them to the same.

client/client.go
ggwpez: Added `setOpTrust`.

eth/client.go
ggwpez: I added a `startBlock *uint64` argument to both subscribe functions.

client/client.go
ggwpez: The commit looks a bit weird but in the end result there is a newline above the `sendTx`.

eth/client.go
ggwpez: I added a startBlock *uint64 argument to both subscribe functions.

sebastianst: We discussed with @matthiasgeihs that subscriptions only for for future blocks anyways. So we need to check if it's sufficient in our use cases to only have subscriptions to events starting at the moment of subscription. We can then remove the `startBlock` because we don't need to only start at some future block.

eth/client.go
sebastianst: We discussed with @matthiasgeihs that subscriptions only for for future blocks anyways. So we need to check if it's sufficient in our use cases to only have subscriptions to events starting at the moment of subscription. We can then remove the `startBlock` because we don't need to only start at some future block.

sebastianst:
Commit a82b91149fe17023f17a03e52c49483e6b6fcc1e has the wrong gitmoji, should be :sparkles: Note that we don't really use gitmoji in Erdstall so you can also just stop using it ^^ I use the more common style `pkg: Do stuff` here.

sebastianst (commented):
Please check the usage of our subscriptions, if they work with normal subs, without past filters. If yes, we can remove the `startBlock`. If no, we have to additionally do filtering into the past. Then also check, if we really want all events since `InitBlock` or actually only of the active epoch.

eth/client.go
ggwpez: I really hope it is generally sufficient, at least in the cases that i tested it worked.

eth/client.go
ggwpez: > Please check the usage of our subscriptions, if they work with normal subs, without past filters. If yes, we can remove the `startBlock`. If no, we have to additionally do filtering into the past. Then also check, if we really want all events since `InitBlock` or actually only of the active epoch.

I tried it three times with three clients both malicious and honest case and it worked. Maybe we could even do a CI test at some point :raised_hands:

ggwpez:
> Commit [a82b911](https://github.com/perun-network/erdstall/commit/a82b91149fe17023f17a03e52c49483e6b6fcc1e) has the wrong gitmoji, should be  Note that we don't really use gitmoji in Erdstall so you can also just stop using it ^^ I use the more common style `pkg: Do stuff` here.

Changed it to :sparkles: now.

sebastianst (approved):

PR #61, opened 2020-11-22
Add WS RPC, Messages and RemoteEnclave

### LINKED ISSUE #41: Use proof subscriptions/blocking call instead of polling in client/operator communication
The client currently polls the operator for fresh deposit and balance proofs. It would be better to either
1. use a subscription instead or
2. make the call blocking, like the enclave call already is.

The issue with 2. might be that it only works on a very stable and uninterrupted connection between client and op. But it would be more straight forward to change the current implementation to that.

ggwpez:
Signed-off-by: Oliver Tale-Yazdi <oliver@perun.network>


wire/message.go
```diff
@@ -0,0 +1,92 @@
...
+		Error string `json:"error,omitempty"`
+	}
+
+	// SendTx sends one Transaction to the remote enclave.
+	SendTx struct {
+		Call
+		Tx tee.Transaction `json:"tx"`
+	}
+
+	// Subscribe subscribes to a topic.
```
sebastianst: "Subsribe sets up a client subscription"
we don't subscribe to individual topics.

tee/test/rng.go
```diff
@@ -0,0 +1,60 @@
...
+func RandomDP(rng *rand.Rand) *tee.DepositProof {
+	return &tee.DepositProof{
+		Balance: RandomBalance(rng),
+		Sig:     RandomSig(rng),
+	}
+}
+
+// RandomSig produces invalid ECDSA signatures, but does not need a wallet
+// to work.
+func RandomSig(rng *rand.Rand) tee.Sig {
```
sebastianst: only need to pass `io.Reader`.

tee/test/rng.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+	"perun.network/go-perun/backend/ethereum/wallet/hd"
+	wtest "perun.network/go-perun/backend/ethereum/wallet/test"
+
+	"github.com/perun-network/erdstall/eth"
+	"github.com/perun-network/erdstall/tee"
+)
+
+func RandomTx(require *require.Assertions, rng *rand.Rand) *tee.Transaction {
```
sebastianst: why don't you, as usual, pass a `t require.TestingT`? Looks strange.

tee/transaction_test.go
```diff
@@ -32,3 +35,22 @@ func TestTransaction_SignVerify(t *testing.T) {
...
+
+func TestTransaction_Json(t *testing.T) {
+	rng := rand.New(rand.NewSource(123))
+	sig := make([]byte, 65)
+	rng.Read(sig)
+
+	sender := ttest.NewRandomAddress(rng)
+	receiver := ttest.NewRandomAddress(rng)
+
+	tx := tee.Transaction{
```
sebastianst: why don't you just use `RandomTx` here but manually assemble a random tx?

client/rpc.go
```diff
@@ -4,70 +4,229 @@ package client
...
+		Embedding: perunlog.MakeEmbedding(perunlog.WithField("role", "client")),
 	}
-	return &RPC{conn}
+	if err := conn.SetReadDeadline(time.Time{}); err != nil {
+		return nil, err
+	}
+	conn.SetCloseHandler(func(code int, err string) error {
+		rpc.Log().WithField("code", code).Errorf("WS connection was closed: %s", err)
+		rpc.Close()
+		return nil
```
sebastianst: do you on purpose not `return rpc.Close()`?

client/rpc.go
```diff
@@ -4,70 +4,229 @@ package client
...
+// SendTx sends one transaction to the operator.
+func (r *RPC) SendTx(ctx context.Context, tx tee.Transaction) error {
+	call := wire.NewSendTx(r.nextID(), tx)
+	errChan := make(chan error)
+	// Setup async response cb.
+	r.registerCallback(call.Call.ID, func(result wire.Result, msg []byte) {
+		if result.Error != "" {
+			errChan <- fmt.Errorf("SendTx RPC result: %s", result.Error)
+		} else {
+			close(errChan)
```
sebastianst: I think it would be more consistent to also just send a `nil` here instead of closing the channel.

client/rpc.go
```diff
@@ -4,70 +4,229 @@ package client
...
+	case err := <-errChan:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// GetDepositProof blocks until it can return the next deposit proof from the
+// operator or an error if the context ran out.
+func (r *RPC) GetDepositProof(ctx context.Context) (tee.DepositProof, error) {
```
sebastianst: Getters usually don't have `Get` in their name in Go. I'd either remove the `Get` or use `Next` as a prefix, then acting more like an iterator. Same for BalanceProofs.

client/rpc.go
```diff
@@ -4,70 +4,229 @@ package client
...
+// `GetBalanceProof`.
+func (r *RPC) Subscribe(ctx context.Context, user common.Address) error {
+	call := wire.NewSubscribe(r.nextID(), user)
+	errChan := make(chan error)
+	// Setup async response cb.
+	r.registerCallback(call.Call.ID, func(result wire.Result, msg []byte) {
+		if result.Error != "" {
+			errChan <- fmt.Errorf("Subscribe RPC result: %s", result.Error)
+		} else {
+			close(errChan)
```
sebastianst: again, I'd just push a `nil` here for consistency.

client/rpc.go
```diff
@@ -4,70 +4,229 @@ package client
...
 }
 
-func (r *RPC) GetDepositProof(ctx context.Context, epoch uint64, user common.Address) (reply tee.DepositProof, err error) {
-	for i := 0; ; i++ {
-		if e := r.call(ctx, "RemoteEnclave.GetDepositProof", user, &reply); e != nil {
-			return tee.DepositProof{}, e
+// Subscribe subscribes to the Balance and Deposit proof topic.
+// The user must always read the proofs from via `GetDepositProof` and
+// `GetBalanceProof`.
+func (r *RPC) Subscribe(ctx context.Context, user common.Address) error {
```
sebastianst: We could also change the API to let `Subscribe` return an object that acts like an iterator for balance and deposit proofs, having the `BalanceProof` and `DepositProof` methods. Then we have less method cluttering on the `RPC`, and possibly also less state in it? Because we can then move the channels into it. Better encapsulation.

client/rpc.go
```diff
@@ -4,70 +4,229 @@ package client
...
 		return
 	}
+	r.callbacks[id] = cb
 }
 
-func (r *RPC) GetBalanceProof(ctx context.Context, user common.Address) (reply tee.BalanceProof, err error) {
-	return reply, r.call(ctx, "RemoteEnclave.GetBalanceProof", user, &reply)
+func (r *RPC) callCallback(result wire.Result, data []byte) {
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
```
sebastianst: You should unlock the mutex already after retrieving the callback.Otherwise the mutex is locked until the callback got executed.

sebastianst: You can also change the mutex to a `RWMutex` and only read here. And I realize that you don't delete callbacks, so this map grows indefinitely. So after calling `cb`, you should write-lock the map again and delete the callback.

client/rpc.go
```diff
@@ -4,70 +4,229 @@ package client
...
-				return nil
-			}
-			err = fmt.Errorf("Retry %d: %s", i, errors.Cause(e))
-			time.Sleep(time.Second * 1)
-		case <-ctx.Done():
-			if err != nil {
-				return err
-			}
-			return ctx.Err()
+func (r *RPC) handleTopic(topic wire.Topic, data []byte) {
```
sebastianst: If you create a subscription class, you can move this method to it.

client/rpc.go
```diff
@@ -4,70 +4,229 @@ package client
...
+		r.Log().WithField("topic", topic).Error("unknown result topic")
 	}
 }
+
+func (r *RPC) nextID() wire.ID {
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+	r.id++
+	return wire.ID(strconv.Itoa(r.id))
+}
```
sebastianst: seeing this, I'm not sure if it makes sense to have the common mutex `mtx` for ids and callbacks. Why not use an atomic `uint64` here instead?

operator/rpc.go
```diff
@@ -1,46 +1,182 @@
...
+	pkgsync "perun.network/go-perun/pkg/sync"
 )
 
-// RemoteEnclave provides the enclave interface to the user.
-type RemoteEnclave struct {
-	operator *Operator
-}
+type (
+	// RPC handels RPC request and forwards them to the enclave.
+	RPC struct {
```
sebastianst: Maybe rename to `RPCServer`, it's unclear what it is

operator/rpcoperator.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/perun-network/erdstall/tee"
+)
+
+type (
+	RemoteOperator interface {
+		AddTransaction(tee.Transaction) error
```
sebastianst: Rename to `Send`

operator/rpcoperator.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+func NewRPCOperator(enclave tee.Enclave) *RPCOperator {
+	return &RPCOperator{
+		enclave: enclave,
+		subs:    make(map[common.Address]subAndInputs),
+	}
+}
+
+func (o *RPCOperator) AddTransaction(tx tee.Transaction) error {
```
sebastianst: Rename to `Send`

operator/rpcoperator.go
```diff
@@ -0,0 +1,151 @@
...
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/perun-network/erdstall/tee"
+)
+
+type (
+	RemoteOperator interface {
```
sebastianst: RPC(Server), RPCOperator, RemoteOperator all gets a bit confusing...
Maybe we call the `RemoteOperator` `WireOperator` or `WireAPI` or `WireInterface` instead?

operator/rpcoperator.go
```diff
@@ -0,0 +1,151 @@
...
+		AddTransaction(tee.Transaction) error
+		SubscribeProofs(common.Address) (ProofSub, error)
+	}
+
+	// RPCOperator is the will be exposed to the client via websocket RPC.
```
sebastianst:
```suggestion
	// RPCOperator will be exposed to the client as a websocket RPC Server.
```

operator/rpcoperator.go
```diff
@@ -0,0 +1,151 @@
...
+		deposits <-chan tee.DepositProof
+		balances <-chan tee.BalanceProof
+		quit     chan struct{}
+	}
+
+	subAndInputs struct {
+		sub  *ProofSub
+		deps chan tee.DepositProof
+		bals chan tee.BalanceProof
+	}
```
sebastianst: From what I can see, you don't need this unnecessary wrapper type `subAndInputs`, just use a `map[common.Address]ProofSub` in the `RPCOperator`. It's enough to expose the channels as read-only channels in the getters `ProofSub.Deposits` and `Balances`. The class fields can just be normal channels in `ProofSub`.

operator/rpcoperator.go
```diff
@@ -0,0 +1,151 @@
...
+	// Write new proof in.
+	select {
+	case <-sub.sub.quit:
+		delete(o.subs, who)
+	case sub.bals <- proof:
+	}
+}
+
+// NewProofSub returns a new proofSub with the given input channels.
+func NewProofSub(deposits <-chan tee.DepositProof, balances <-chan tee.BalanceProof) *ProofSub {
```
sebastianst: I think we don't need to export this constructor. The user of the `operator` package should never create subs, right?

operator/rpc.go
```diff
@@ -1,46 +1,182 @@
...
 }
+
+func (p *Peer) sendResult(id wire.ID, _err error) {
+	err := ""
+	if _err != nil {
+		err = _err.Error()
+	}
+	response := wire.Result{ID: id, Error: err}
+	if err := p.sendJSON(response); err != nil {
+		p.Log().WithError(err).Error("Could not send message.")
```
sebastianst: Why do you handle the error here and not return it to the caller? You call it in `readMessage` where you also handle the error of `ReadMessage`.

operator/rpc.go
```diff
@@ -1,46 +1,182 @@
...
+	}
+}
+
+func (p *Peer) sendJSON(obj interface{}) error {
+	p.connMtx.Lock()
+	defer p.connMtx.Unlock()
+	return p.conn.WriteJSON(obj)
```
sebastianst, line 185: So websocket conns are definitely not thread safe?

operator/rpc.go
```diff
@@ -1,46 +1,182 @@
...
+				}
+			case proof := <-p.sub.balances:
+				update = &wire.BalanceProof{
+					Result: wire.Result{
+						Topic: wire.BalanceProofs,
+					},
+					Proof: proof,
+				}
+			case <-p.sub.quit:
+				break
```
sebastianst: Add logging that the subscription routine returns due to a closed sub.

operator/test/rpcoperator.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type (
+	// RPCOperator is a mocked operator.RPCOperator.
+	// It can be used as an RPCOperator for testing the operators
+	// RPC module, since it expects a operator.RPCOperator.
+	//
+	// There are two functions for setting errors that should be returned
+	// when the mocked functions are called.
+	RPCOperator struct {
```
sebastianst, line 21: if we rename the `operator.RPCOperator` we should do it similarly here.

operator/test/rpcoperator.go
```diff
@@ -0,0 +1,82 @@
...
+		op:      op.NewRPCOperator(enclave),
+	}
+}
+
+// Run pulls the proofs from the enclave and write them to the subscriptions.
+// This is normally done by the Operator. Returns immediately.
+func (r *RPCOperator) Run() {
+	go func() {
+		for {
+			proofs, _ := r.enclave.DepositProofs()
```
sebastianst, line 45: Why don't you check the error, log it if not nil and return?

operator/test/rpcoperator.go
```diff
@@ -0,0 +1,82 @@
...
+			for _, proof := range proofs {
+				r.op.PushDepositProof(*proof)
+			}
+			time.Sleep(100 * time.Millisecond) // poll interval
+		}
+	}()
+	go func() {
+		for {
+			proofs, _ := r.enclave.BalanceProofs()
```
sebastianst, line 55: Why don't you check the error, log it if not nil and return?

operator/test/rpcoperator.go
```diff
@@ -0,0 +1,82 @@
...
+			time.Sleep(100 * time.Millisecond) // poll interval
+		}
+	}()
+}
+
+// AddTransaction is part of the operator.RemoteEnclave interface and adds
+// a Transaction to the enclave.
+// Can be read back from Transactions() which bufferes one TX.
+// Returns the error that was set by SetAddTransactionError.
+func (r *RPCOperator) AddTransaction(tx tee.Transaction) error {
```
sebastianst: Rename to `Send`...

operator/test/enclave.go
```diff
@@ -0,0 +1,96 @@
...
+		case proof := <-e.bals:
+			ret = append(ret, proof)
+		default:
+			return
+		}
+	}
+}
+
+func (e *Enclave) Shutdown() {
+}
```
sebastianst: remove newline

client/rpc_test.go
```diff
@@ -0,0 +1,122 @@
...
+// Client <-> RPC <-> MockedRPCOperator <-> OP
+func TestRPC_ClientOp(t *testing.T) {
+	plogrus.Set(logrus.TraceLevel, &logrus.TextFormatter{ForceColors: true})
+	rng := pkgtest.Prng(t)
+	enclave := optest.NewMockedEnclave()
+	op := optest.NewRPROperator(enclave)
+	op.Run()
+	rpcServer := operator.NewRPC(op)
+	go func() {
+		if err := rpcServer.Serve("", 8401); err != nil {
```
sebastianst: Put port in `const`

client/rpc_test.go
```diff
@@ -0,0 +1,122 @@
...
+			panic(err)
+		}
+	}()
+
+	time.Sleep(shortWait)
+	client, err := client.NewRPC("0.0.0.0", 8401)
```
sebastianst: use port const here

client/rpc_test.go
```diff
@@ -0,0 +1,122 @@
...
+var longWait = 10 * time.Second
+
+// TestRPC_ClientOp is a test which tests all functions of the Operator
+// over a websocket connection.
+//
+// It uses a mocked remote enclave to simulate failing functions.
+// The construction looks a bit like this:
+// Client <-> RPC <-> MockedRPCOperator <-> OP
+func TestRPC_ClientOp(t *testing.T) {
+	plogrus.Set(logrus.TraceLevel, &logrus.TextFormatter{ForceColors: true})
```
sebastianst: do we really always want trace logging in ci tests? I'd rather set this to `Info` and only use trace during debugging.

client/rpc_test.go
```diff
@@ -0,0 +1,122 @@
...
+		// Push in proofs for other clients.
+		enclave.PushBalanceProof(ttest.RandomBP(rng))
+		enclave.PushDepositProof(ttest.RandomDP(rng))
+	}
+
+	t.Run("Subscribe-error", func(t *testing.T) {
+		op.SetSubscribeProofsError(myErr)
+		assert.Error(t, client.Subscribe(ctx, user1))
+		op.SetSubscribeProofsError(nil)
+	})
```
sebastianst, line 91: better add newlines after `t.Run` calls

sebastianst (changes requested):

tee/test/rng.go
ggwpez: It only needed the require, but i can change that.

client/rpc.go
ggwpez: Oh right, this was a remnant from the the old code.

client/rpc.go
ggwpez: Did not think about it :laughing:   
PS: Done

operator/rpcoperator.go
ggwpez: Yes i was having trouble to find good names, will try em out.  
PS: Used `WireAPI` now

operator/rpc.go
ggwpez: The gorilla implementation of them is not, the race detector is unhappy without the mutex.

tee/transaction_test.go
ggwpez: Good point

client/rpc.go
ggwpez: Something with the linters must be wrong, they should detect this... i changed it

client/rpc.go
ggwpez: I will try it out  
PS: Done

client/rpc.go
ggwpez: Right, i added the delete logic

operator/rpcoperator.go
ggwpez: Ok thanks for the heads up, i investigated the issue again and could remove it.  
I already tried it before but failed, the solution was to expose the `quit` channel by a `Closed()` getter.

operator/rpc.go
ggwpez: Well ok, does not make a big difference imho, i changed it.

operator/test/rpcoperator.go
ggwpez: We renamed the `RemoteOperator` interface to `WireAPI`, the implementing `RPCOperator` keeps its name, unless you also want to change that.

operator/test/rpcoperator.go
ggwpez: The subscriptions returned by the mocked enclave never return errors, so i omitted a check here (but added a comment :laughing:).

operator/rpc.go
ggwpez: Added `p.Log().Debug("Subscription routine returns due to closed sub.")`.

client/rpc_test.go
ggwpez: Added `opRPCPort`.

client/rpc_test.go
ggwpez: Removed it.

tee/tee.go
```diff
@@ -135,38 +137,26 @@ func (s *Sig) UnmarshalJSON(data []byte) error {
...
 }
 
-func (b Balance) MarshalJSON() ([]byte, error) {
-	aux := map[string]interface{}{
-		"epoch":   b.Epoch,
-		"account": b.Account,
-		"value":   b.Value.Text(10),
-	}
-	return json.Marshal(aux)
+func (s Amount) MarshalJSON() ([]byte, error) {
```
sebastianst: Use `a` as var for all `Amount` methods and pointer.
```suggestion
func (s *Amount) MarshalJSON() ([]byte, error) {
```

operator/test/user.go
```diff
@@ -165,8 +165,9 @@ func (u *User) BalanceProof(ctx context.Context) {
 	}
 	u.bp = proof
 
-	if u.bp.Balance.Value.Int64() != u.TargetBalance {
-		u.Errorf("incorrect balance, got %d, expected %d", u.bp.Balance.Value.Int64(), u.TargetBalance)
+	balance := (*big.Int)(u.bp.Balance.Value).Int64()
+	if balance != u.TargetBalance {
```
sebastianst: merge

operator/rpcoperator.go
```diff
@@ -0,0 +1,148 @@
...
+	return *sub, nil
+}
+
+// subscribe is an internal implementation detail and should not be called.
+func (o *RPCOperator) subscribe(addr common.Address) ProofSub {
+	o.subs[addr] = newProofSub()
+	return *o.subs[addr]
+}
+
+func (o *RPCOperator) PushDepositProof(proof tee.DepositProof) {
```
sebastianst, line 75: Do we really need to export the `Push...` methods? I mean, a subscriber outside this package isn't pushing new proofs into the sub, right? Same for balance proofs.

client/rpc.go
```diff
@@ -4,70 +4,260 @@ package client
...
 		}
+
+		switch {
+		case msg.ID != "":
+			r.callCallback(msg, data)
+		case msg.Topic != "":
+			r.subMtx.RLock()
+			// Don't unlock here, since otherwise a `Subscribe` call could
+			// override the subscription while it receives a proof.
+			defer r.subMtx.RUnlock()
```
sebastianst: Never call `defer`s inside loops. What you're doing here is acquiring indefinitely many read locks and stack indefinitely many unlocks, all executed at once when the `handleConnections` go routine returns.

Also I'm not sure if we even need the `subMtx`. I'll leave a comment about this somewhere else.

client/rpc.go
```diff
@@ -4,70 +4,260 @@ package client
...
-func (r *RPC) GetDepositProof(ctx context.Context, epoch uint64, user common.Address) (reply tee.DepositProof, err error) {
-	for i := 0; ; i++ {
-		if e := r.call(ctx, "RemoteEnclave.GetDepositProof", user, &reply); e != nil {
-			return tee.DepositProof{}, e
+// Subscribe subscribes to the Balance and Deposit proof topic.
+// The user must always read the proofs from via `DepositProof` and
+// `BalanceProof`.
+func (r *RPC) Subscribe(ctx context.Context, user common.Address) (*Subscription, error) {
+	r.subMtx.Lock()
+	r.subscription = &Subscription{
```
sebastianst, line 111: Do we even need the `subMtx`? It looks to me that `RPC.Subscribe` is called once during setup of the client and from then on is used, but never reset, no? Is it even valid to call `Subscribe` twice?

client/rpc.go
```diff
@@ -4,70 +4,260 @@ package client
...
+		perunlog.Embedding
+
+		connMtx sync.Mutex // protects conn.send.
+		conn    *gorilla.Conn
+
+		cbMtx     sync.RWMutex // protects callbacks.
+		id        uint64
+		callbacks map[wire.ID]callback
+
+		subMtx       sync.RWMutex // protects subscription.
```
sebastianst: Not sure if even needed, see other comment.

sebastianst (changes requested):

operator/rpcoperator.go
ggwpez: It is used in a Black Box test and must therefore sadly be exported.

client/rpc.go
ggwpez: 🤦

client/rpc.go
ggwpez: Removed it since it would only theoretically be needed in testing code.

operator/test/user.go
ggwpez: It is used again one line below, i thought it looks better to only have the `(*big.Int)(u.bp.Balance.Value).Int64()` once.

sebastianst:
```suggestion
	if balance := (*big.Int)(u.bp.Balance.Value).Int64(); balance != u.TargetBalance {
```

client/rpc.go
ggwpez: The client assumes it to be valid to subscribe twice, because otherwise that case could not be tested.  
The RPC Server rejects a double subscription obviously.

ggwpez: But i guess since we only call it twice in testing code, i will remove it.

operator/test/user.go
ggwpez: Ok i did not get it^^

tee/tee.go
ggwpez: PS The marshall needs an object and not a pointer, it silently fails otherwise.

client/rpc.go
```diff
@@ -4,70 +4,254 @@ package client
...
+			r.Log().Error("Received result without ID or Topic")
+		}
+	}
+}
+
+func (r *RPC) registerCallback(id wire.ID, cb callback) {
+	r.cbMtx.Lock()
+	defer r.cbMtx.Unlock()
+	if _, ok := r.callbacks[id]; ok {
+		r.Log().Error("Callbacks should never be overridden.")
```
sebastianst: "overwritten"

Since you're writing an error to the log here and not return an error, this should be rephrased to a sentence like
```suggestion
		r.Log().WithField("id", id).Error("Callback already registered, skipping.")
```

sebastianst (changes requested):

sebastianst (approved):

PR #62, opened 2020-12-01
Recover deposit

### LINKED ISSUE #51: Extract and log transaction failure reason
The operator tracks the status of challenge response transactions and logs an error if a transaction fails. Currently we do not log the failure reason because it is unclear how failure information can be extracted from go-ethereum.

Ideally we would like to log an error with the error reason (e.g., "out of gas") if a challenge response transaction fails.

### LINKED ISSUE #54: Use ContractBackend.ConfirmTransaction instead of abi.WaitMined where possible
Has the additional benefit of determining revert reasons.

ggwpez:


sebastianst:
Note that we should first merge #66 where the contract usage will change. There's now only one `withdrawFrozen` function but two different challenge functions.

ggwpez:
Done in #70.

PR #66, opened 2020-12-07
contract: Fix challenge logic

### LINKED ISSUE #57: Contract: make freezing idempotent and/or implicit
Right now, at least one participant has to call `freeze` once the freezing phase starts. If there are multiple clients that want to freeze the contract, their client impls would probably all call `freeze` at the same time, resulting in many reverting and unnecessary transactions.

### LINKED ISSUE #65: Contract: Only system participants should be able to challenge operator

sebastianst:
Resolves #57, resolves #65


sebastianst:
Forgot to regenerate bindings :joy: Will only remove the `freeze` call though as the remaining API is unchanged.

I am also surprised that the tests work, don't we test freezing somewhere?

contracts/contracts/Erdstall.sol
```diff
@@ -190,6 +199,10 @@ contract Erdstall {
         return epoch()-3;
     }
 
+    function sealedEpoch() internal view returns (uint64) {
```
ggwpez, line 227: I will add a `SealedEpoch(uint64)` to `params.go`, you can then later-on merge it into this commit.

contracts/contracts/Erdstall.sol
```diff
@@ -101,63 +99,99 @@ contract Erdstall {
...
         numChallenges[epoch]++;
 
         emit Challenged(epoch, msg.sender);
     }
 
-    // Freezes the contract such that only exits of the prior exit phase can be
-    // posted.
-    // Can be called by any user if the prior phase has an open challenge.
-    // This must be called in the phase following the challenged exit phase.
-    function freeze() external {
```
matthiasgeihs, line 122: What do you think about keeping a method `freeze` for convenience and clarity? It's essential that `withdrawChallenge` is called in the immediately following epoch after the challenged epoch, right? With withdrawals you could usually also wait until later periods, right? With the current naming, somebody could think he can also just withdraw later, but then it's too late.

contracts/contracts/Erdstall.sol
```diff
@@ -101,63 +99,99 @@ contract Erdstall {
...
-        require(!isFrozen(), "already frozen");
-        require(isLastEpochChallenged(), "no challenge in last epoch");
+    // withdrawChallenge lets open challengers withdraw all funds locked in the
+    // frozen contract. The funds were already determined when the challenge was
+    // posted using either `challenge` or `challengeDeposit`.
+    //
+    // Implicitly calls ensureFrozen to ensure that the contract state is set to
+    // frozen if the last epoch has an unanswered challenge.
+    function withdrawChallenge() external {
```
matthiasgeihs, line 145: In the interest of simplicity, what it be possible to remove `withdrawChallenge` and only use `withdrawFrozen`?

contracts/contracts/Erdstall.sol
```diff
@@ -190,6 +224,10 @@ contract Erdstall {
         return epoch()-3;
     }
 
+    function sealedEpoch() internal view returns (uint64) {
```
matthiasgeihs, line 227: Seems redundant to `freezingEpoch`?

matthiasgeihs (commented):
Had a brief look at the contract. Generally seems ok to me, just a few comments.

contracts/contracts/Erdstall.sol
sebastianst: It's on purpose. Semantically clearer.

contracts/contracts/Erdstall.sol
sebastianst: It would be and I thought about it, but it would be less optimal because then passing the proof and verifying it would happen twice, which is quite expensive.

contracts/contracts/Erdstall.sol
sebastianst: Already there, now called `ensureFrozen public`. Note that `public` is both, `internal` and `external`.

tee/tee.go
```diff
@@ -196,3 +198,11 @@ func (b *Block) GobDecode(blob []byte) error {
 	}
 	return nil
 }
+
+func (b *Balance) ToEthBals() bindings.ErdstallBalance {
```
sebastianst, line 202: I'd name it `ToEthBal` or `ToEthBalance` since it's singular

operator/operator.go
```diff
@@ -241,12 +241,7 @@ func (operator *Operator) handleChallengedEvent(c challengedEvent) error {
...
 	}
 
-	balance := bindings.ErdstallBalance{
-		Epoch:   balanceProof.Balance.Epoch,
-		Account: balanceProof.Balance.Account,
-		Value:   (*big.Int)(balanceProof.Balance.Value),
-	}
-
+	balance := balanceProof.Balance.ToEthBals()
 	tx, err := operator.contract.Exit(tr, balance, balanceProof.Sig)
```
sebastianst, line 245: You could move it directly into the argument since `balance` is only moved once.

client/client.go
```diff
@@ -599,7 +633,10 @@ func (c *Client) listenOnChain() error {
 		subError <- c.ethClient.SubscribeEpochs(c.Ctx(), *c.params, epochs, blocks)
 	}()
 	go c.BalanceProofWatcher()
-	go c.FrozenWatcher()
+	frozenCtx, cancel := context.WithCancel(c.Ctx())
+	c.stopFrozenWatcher = cancel
+	go c.FrozenWatcher(frozenCtx)
```
sebastianst, line 638: Since you start the frozen watcher from this listener, do we really want to make the `FrozenWatcher` method public?

client/client.go
```diff
@@ -439,14 +431,26 @@ func (c *Client) handleFrozen(event *bindings.ErdstallFrozen) {
 			}
 		}
 	}()
+	c.withdrawFrozen(status, epoch)
+}
+
+func (c *Client) withdrawFrozen(status chan *CmdStatus, epoch uint64) {
+	c.balMtx.Lock()
+	defer c.balMtx.Unlock()
```
sebastianst, line 439: Do you really need to lock the mutex for the whole duration of the function? And if it's a `RWLock` a `RLock` would be sufficient.

client/client.go
```diff
@@ -494,26 +498,55 @@ func (c *Client) CmdChallenge(status chan *CmdStatus, args ...string) {
...
 		if err := c.ethClient.WaitForBlock(c.Ctx(), next); err != nil {
 			status <- &CmdStatus{Err: err}
 			return
 		}
 	}
 
+	// Get the BP of the sealed epoch.
+	lastBlock = atomic.LoadUint64(&c.lastBlock) + 1
+	sealed := c.params.SealedEpoch(lastBlock)
+	c.balMtx.Lock()
```
sebastianst, line 511: RLock?

client/client.go
```diff
@@ -494,26 +498,55 @@ func (c *Client) CmdChallenge(status chan *CmdStatus, args ...string) {
...
-		next := c.params.DepositDoneBlock(c.params.ExitEpoch(lastBlock + 1))
+		next := c.params.DepositDoneBlock(c.params.DepositEpoch(lastBlock + 1))
 		if err := c.ethClient.WaitForBlock(c.Ctx(), next); err != nil {
 			status <- &CmdStatus{Err: err}
 			return
 		}
 	}
 
+	// Get the BP of the sealed epoch.
+	lastBlock = atomic.LoadUint64(&c.lastBlock) + 1
```
sebastianst, line 509: It would be nice to put all `atomic.LoadUint64(&c.lastBlock)` into its own getter `LastBlock`. Maybe also add another getter for the +1 version called `ActiveBlock`.

client/client.go
```diff
@@ -494,26 +498,55 @@ func (c *Client) CmdChallenge(status chan *CmdStatus, args ...string) {
...
+		return c.contract.Challenge(auth, bp.Balance.ToEthBals(), bp.Sig)
 	}, status)
 	if err != nil {
 		status <- &CmdStatus{Err: fmt.Errorf("Challenge TX: %w", err)}
 		return
 	}
+	c.waitForReponse(status, tx.BlockNumber.Uint64())
+}
 
+func (c *Client) waitForReponse(status chan *CmdStatus, block uint64) {
```
sebastianst, line 544:
```suggestion
func (c *Client) waitForResponseOrWithdraw(status chan *CmdStatus, block uint64) {
```
there was a missing `s` in `Response` and since this function is actually also doing something (withdrawing), I'd state this explicitly in its name.

client/client.go
```diff
@@ -397,8 +393,8 @@ func (c *Client) BalanceProofWatcher() {
 
 // FrozenWatcher listens for Frozen events and calls WithdrawFrozen
 // if a balance proof is available.
-func (c *Client) FrozenWatcher() {
-	ctx, cancel := context.WithCancel(c.Ctx())
+func (c *Client) FrozenWatcher(_ctx context.Context) {
+	ctx, cancel := context.WithCancel(_ctx)
 	defer cancel()
```
sebastianst, line 398: you already pass a cancel-context whose cancel function you set as the watcher canceller. So you don't need this cancel wrapper.

But I think it's actually better to not pass a context to the `FrozenWatcher` at all and set
`ctx, c.stopFrozenWatcher = context.WithCancel(c.Ctx())` here directly.

contracts/contracts/Erdstall.sol
ggwpez: This comment was still in my review buffer, outdated now.

client/client.go
ggwpez: Yes good idea

client/client.go
ggwpez: \+ added a comment

client/client.go
ggwpez: private now

client/client.go
ggwpez: fixed

client/client.go
ggwpez: Added both

PR #68, opened 2020-12-08
Fix epoch merge

ndzik:
Our Epoch-Merge method did not act on the correct balance. Thus we never applied changes that occurred due to extra deposits.


sebastianst:
We still need to write a test for this. Next time we touch the enclave test...

PR #73, opened 2020-12-17
Enclave Caching

### LINKED ISSUE #64: Add TX Caching to Operator
Currently, the operator forwards incoming transactions straight to the enclave. This is unstable near phase shifts, where the operator might receive TXs for the next epoch already. We therefore need to enable tx caching for incoming transactions. Once a phase shift is performed, all cached transactions for this epoch should be pushed into the enclave.

sebastianst:
Resolves #64


sebastianst:
183df8a4 operator handle msg in go routine commit

sebastianst:
Obsolete due to phase merge in `erdstall-ext` and because we might remove the `Epoch` field altogether from Txs.

PR #74, opened 2020-12-17
[operator] Fix rpc subscription

ndzik:
The operator did not properly unsubscribe subscriptions, when a
connection was closed. Proof-Channels for subscriptions were
shared between multiple peer connections, thus resulting in pushing
proofs in a round-robin fashion.

We now multiplex proofs over all existing subscriptions.

### DIFF
--- a/operator/rpcoperator.go
+++ b/operator/rpcoperator.go
@@ -27,7 +27,15 @@ type (
 		mtx     sync.Mutex // protects all
 		enclave tee.Enclave
 
-		subs map[common.Address]*ProofSub
+		subs map[common.Address]*BufferedProofSubs
+	}
+
+	// BufferedProofSub describes a slice of subs, which also holds the latest
+	// available deposit- and/or balance-proof.
+	BufferedProofSubs struct {
+		subs     []ProofSub
+		latestDP *tee.DepositProof
+		latestBP *tee.BalanceProof
 	}
 
 	// ProofSub is a subscription on TEE proofs.
@@ -43,7 +51,7 @@ var _ WireAPI = (*RPCOperator)(nil)
 func NewRPCOperator(enclave tee.Enclave) *RPCOperator {
 	return &RPCOperator{
 		enclave: enclave,
-		subs:    make(map[common.Address]*ProofSub),
+		subs:    make(map[common.Address]*BufferedProofSubs),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


operator/rpcoperator.go
```diff
@@ -60,63 +68,86 @@ func (o *RPCOperator) SubscribeProofs(addr common.Address) (ProofSub, error) {
...
+		// Write new proof in.
+		select {
+		case <-subs[i].quit:
+			l := len(subs) - 1
+			subs[i] = subs[l]
+			subs = subs[:l]
+			i--
+		case subs[i].deposits <- proof:
+		}
 	}
```
sebastianst, line 120: Not sure, don't you need to set `bufsub.sub = subs` here? Otherwise the cleanup won't be persistent. Same for BPs.

operator/rpcoperator.go
```diff
@@ -60,63 +68,86 @@ func (o *RPCOperator) SubscribeProofs(addr common.Address) (ProofSub, error) {
...
-	select {
-	case <-sub.deposits:
-	default:
-	}
-	// Write new proof in.
-	select {
-	case <-sub.quit:
-		delete(o.subs, who)
-	case sub.deposits <- proof:
+	o.subs[who].latestDP = &proof
```
sebastianst: `bufsub` is a pointer, so you could do
```suggestion
	bufsub.latestDP = &proof
```
here, no?

operator/rpcoperator.go
```diff
@@ -60,63 +68,86 @@ func (o *RPCOperator) SubscribeProofs(addr common.Address) (ProofSub, error) {
...
-	select {
-	case <-sub.balances:
-	default:
-	}
-	// Write new proof in.
-	select {
-	case <-sub.quit:
-		delete(o.subs, who)
-	case sub.balances <- proof:
+	bufsub.latestBP = &proof
```
sebastianst, line 134: here you do it

operator/rpcoperator.go
```diff
@@ -129,19 +160,19 @@ func newProofSub() *ProofSub {
 	}
 }
 
-func (sub *ProofSub) Deposits() <-chan tee.DepositProof {
+func (sub ProofSub) Deposits() <-chan tee.DepositProof {
```
sebastianst, line 165: why did you switch to value?

sebastianst (changes requested):

operator/rpcoperator.go
ndzik: Channel-references within the `ProofSub` don't change and are readonly, unless you view pushing into a channel as a mutable action. Imo this communicates the intent better. But it is minor, if there are any downsides to this, tell me (:

Performancewise this should not change much, since we pass pointer around anyway.

operator/rpcoperator.go
ndzik: Good point.

operator/rpcoperator.go
ndzik: Very good catch :muscle: 

I will do it like you suggested and won't start doing this via dereferencing the slice, ok?

PR #79, opened 2020-12-22
Add ClientConfiguration Push upon connect

ndzik:
Instead of an elaborate handshake protocol, we settled for a simple `ClientConfig` which is pushed upon establishing a connection with the client. This `ClientConfig` includes the `NetworkID` to which the Operator is connected and `Address` of the corresponding contract.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,14 @@ ganache-cli -e 100000000000000 -b 5 -m "pistol kiwi shrug future ozone ostrich m
 --account-index 3
 ```
 
+### Configuration file
+
+With the command line flag `-config <file>`, you can specify a JSON
+configuration file. In this file, you can override the default settings. For
+reference, see `demo/config.json`. If you wish to enable TLS support for the
+websocket connection, set the fields `KeyFile` to the TLS private key file path,
+and `CertFile` to the TLS certificate file path.
+
 ## Description
 
 Erdstall leverages Trusted Execution Environments (TEE) like Intel SGX (or even
--- a/demo/config.json
+++ b/demo/config.json
@@ -8,8 +8,6 @@
 	"PowDepth": 0,
 	"RpcPort": 8401,
 	"RpcHost": "0.0.0.0",
-        "KeyFile": "/etc/letsencrypt/live/demo.erdstall.dev/privkey.pem",
-        "CertFile": "/etc/letencrypt/live/demo.erdstall.dev/cert.pem",
 	"RespondChallenges": true,
 	"SendDepositProofs": true,
 	"SendBalanceProofs": true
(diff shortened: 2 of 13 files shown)


demo/config.json
```diff
@@ -8,6 +8,8 @@
 	"PowDepth": 0,
 	"RpcPort": 8401,
 	"RpcHost": "0.0.0.0",
+        "KeyFile": "/etc/letsencrypt/live/demo.erdstall.dev/privkey.pem",
```
ggwpez: I think it would be cleaner to create a certificate for the `operator` instead of copying it from the `demo` server.

operator/operator.go
```diff
@@ -172,9 +172,16 @@ func (operator *Operator) Serve(port uint16) error {
 
 	operator.rpcOperator = NewRPCOperator(operator.enclave)
 
+	ctxReq, cancel := eth.ContextNodeReq()
+	defer cancel()
```
sebastianst: You could move this into the new `NetworkID` method and remove the context from its parameters.

demo/config.json
sebastianst: I actually think we shouldn't set TLS paths in the `demo/config` file, but document the json fields. This is a path not relevant for this public repo.

README.md
```diff
@@ -12,6 +12,10 @@ folder. All TEE related code can be found in `tee`.
 
 ## Getting started
 
+If you wish to enable TLS support for the websocket connection, go to
+`demo/config.json`, and set the fields `KeyFile` to the TLS private key file
+path, and `CertFile` to the TLS certificate file path.
```
sebastianst:
```suggestion
If you wish to enable TLS support for the websocket connection, set the fields `KeyFile` to the TLS private key file
path, and `CertFile` to the TLS certificate file path in your configuration json file.
```
This is a general documentation about the configuration file, not specific to file `demo/config.json`. Maybe also put it in a separate sub section further down `### Configuration file`. Also mention that the configuration file can be specified for the op with cmdline flag `-config config.json`.

sebastianst (changes requested):

sebastianst (dismissed):

ggwpez (changes requested):
The failing CI seems to be indeed a bug introduced in this MR.

ggwpez (dismissed):
I dont review it but just wanted to unblock the MR.

operator/rpc.go
```diff
@@ -77,6 +137,10 @@ func (r *RPCServer) connectionHandler(out http.ResponseWriter, in *http.Request)
 	}
 
 	peer := &Peer{conn: conn, op: r.op}
+	if err := peer.sendJSON(r.server.clientConfig); err != nil {
+		r.Log().WithError(err).Error("Pushing ClientConfig:", err)
```
ggwpez: You are logging the error twice. Just looked at it because i have to rebase onto it.

ndzik: Good point, we also did the same mistake five lines above :laughing:

RmbRT (approved):

PR #89, opened 2021-01-13
84 pending tx receipts

### LINKED ISSUE #69: Notify TX recipients
We should let the operator forward incoming transaction to a recipient.
For this we add another subscription topic to the client subscription and just send any incoming transaction unchanged to the receiver.

This message fulfills no protocol purpose but is an informative message for clients. Until now, tx recipients only receive new balance proofs but don't know what txs aggregated to this new balance.

### LINKED ISSUE #84: Pending tx receipts
The operator should forward incoming txs to the recipient, so that their client software can display an incoming pending tx.

We need to add a new subscription message type for this.

Related https://github.com/perun-network/erdstall-web3-client/issues/14

RmbRT:
Closes #84 
Closes #69

### DIFF
--- a/client/rpc.go
+++ b/client/rpc.go
@@ -42,7 +42,8 @@ type (
 		// deposit proofs from the OP will be written into this channel.
 		depProofs chan tee.DepositProof
 		// balance proofs from the OP will be written into this channel.
-		balProofs chan tee.BalanceProof
+		balProofs  chan tee.BalanceProof
+		txReceipts chan tee.Transaction
 	}
 
 	// callback is used for handling call results.
@@ -116,8 +117,9 @@ func (r *RPC) Subscribe(ctx context.Context, user common.Address) (*Subscription
 		// immediately to prevent that they get reordered by a race condition
 		// from the go-routines writing them to the channel since a mutex is
 		// not FIFO.
-		balProofs: make(chan tee.BalanceProof, 10),
-		depProofs: make(chan tee.DepositProof, 10),
+		balProofs:  make(chan tee.BalanceProof, 10),
+		depProofs:  make(chan tee.DepositProof, 10),
+		txReceipts: make(chan tee.Transaction, 10),
 	}
 
 	call := wire.NewSubscribe(r.nextID(), user)
@@ -233,13 +235,28 @@ func (s *Subscription) handleTopic(topic wire.Topic, data []byte) {
 		}
 		s.balProofs <- msg.Proof
 		s.Log().WithField("epoch", msg.Proof.Balance.Epoch).Trace("Received balance proof")
+	case wire.TXReceipts:
+		var msg wire.TXReceipt
+		if err := json.Unmarshal(data, &msg); err != nil {
+			s.Log().WithError(err).Error("decoding json")
+			return
+		}
+		txLog := s.Log().
+			WithField("epoch", msg.TX.Epoch).
+			WithField("amount", msg.TX.Amount)
+		select {
+		case s.txReceipts <- msg.TX:
+			txLog.Debug("Received TX receipt")
+		default:
+			txLog.Debug("Discarded TX receipt")
+		}
 	default:
 		s.Log().WithField("topic", topic).Error("unknown result topic")
 	}
 }
 
 // DepositProof blocks until it can return the next deposit proof from the
-// operator or an error if the context ran out.
+// operator.
 func (s *Subscription) DepositProof(ctx context.Context) (tee.DepositProof, error) {
 	select {
 	case <-ctx.Done():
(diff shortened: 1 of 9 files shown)


operator/operator.go
```diff
@@ -28,6 +28,7 @@ type Operator struct {
 	EthClient *eth.Client
 	*depositProofs
 	*balanceProofs
+	*TxReceipts
```
sebastianst: Why are those embedded? Does `TxReceipts` need to be public?

operator/rpcoperator.go
```diff
@@ -42,16 +43,21 @@ type (
 	ProofSub struct {
 		deposits chan tee.DepositProof
 		balances chan tee.BalanceProof
+		receipts chan tee.Transaction
```
sebastianst, line 47: Don't we also need to buffer receipts in `BufferedProofSub`?

Now that the subscription is not only over proofs but also receipts, we might want to rename `{Buffered,}ProofSub` to just `{Buffered,}Sub` or `...ClientSub`.

operator/rpcoperator.go
```diff
@@ -65,7 +66,22 @@ func (o *RPCOperator) Send(tx tee.Transaction) error {
 	o.mtx.Lock()
 	defer o.mtx.Unlock()
 	log.Infof("Sending %d WEI 0x%s…->0x%s…", (*big.Int)(tx.Amount).Uint64(), tx.Sender.Hex()[:5], tx.Recipient.Hex()[:5])
-	return o.enclave.ProcessTXs(&tx)
+	if err := o.enclave.ProcessTXs(&tx); err != nil {
+		return err
+	}
+	if bsub, ok := o.subs[tx.Recipient]; ok {
+		timeout := time.After(20 * time.Second)
```
sebastianst: I'd add a `const txReceiptDeliveryTimeout = 20 * time.Second` above and use it here.

operator/rpcoperator.go
```diff
@@ -65,7 +66,22 @@ func (o *RPCOperator) Send(tx tee.Transaction) error {
...
+		for _, sub := range bsub.subs {
```
sebastianst, line 76: why can there be multiple subs per recipient?

client/rpc.go
```diff
@@ -234,6 +236,21 @@ func (s *Subscription) handleTopic(topic wire.Topic, data []byte) {
...
+		if err := json.Unmarshal(data, &msg); err != nil {
+			s.Log().WithError(err).Error("decoding json")
+			return
+		}
+		txLog := s.Log().
+			WithField("epoch", msg.TX.Epoch).
+			WithField("amount", msg.TX.Amount)
+		select {
+		case s.txReceipts <- msg.TX:
+			txLog.Trace("Received TX receipt")
```
sebastianst: I think level `Debug` is more appropriate

client/rpc.go
```diff
@@ -260,3 +277,14 @@ func (s *Subscription) BalanceProof(ctx context.Context) (tee.BalanceProof, erro
 		return proof, nil
 	}
 }
+
+// TxReceipt blocks until it can return the next pending incoming transaction
+// from the operator or an error if the context ran out.
```
sebastianst: I find "or an error if the context ran out." redundant. We'd need to write it everywhere, every time, otherwise. It's usual behavior.

operator/operator_test.go
```diff
@@ -53,6 +53,7 @@ func TestOperator(t *testing.T) {
 	// transfer from user1 to user2
 	user1.Transfer(ctx, user2, 3)
 	log.Info("operator_test.TestOperator: Transfer from user1 to user2")
+	user2.TxReceipt(ctx)
```
sebastianst: maybe at least assert the correct amount? same below

sebastianst (changes requested):

operator/operator.go
RmbRT: yes, needs to be public because the operator test needs to access it. I was just following the pattern used for the deposit proofs and balance proofs.

operator/rpcoperator.go
RmbRT: I will rename that, but the receipts need no buffering, since they are only for transactions that occur while someone is connected.

operator/rpcoperator.go
RmbRT: I thought that would be the case ;)

operator/rpcoperator.go
RmbRT: because connections don't get terminated properly and you have multiple connections per peer often times.

operator/rpcoperator.go
sebastianst: OK makes sense, no buffering then :+1: thanks for renaming

operator/operator.go
sebastianst: OK... I think none of them should be embedded but just fields instead. Maybe at least make this a field, can have the same name as the type.

sebastianst (approved):

PR #90, opened 2021-01-14
83 client cfg push

### LINKED ISSUE #83: [client] Handle Client-Config push
Due to the operator now sending a `client-config` over the wire upon a client establishes a connection, we have to adapt the go-client to handle that case.

RmbRT:
Closes #83

### DIFF
--- a/client/rpc.go
+++ b/client/rpc.go
@@ -17,6 +17,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	pkgsync "perun.network/go-perun/pkg/sync"
 
+	"github.com/perun-network/erdstall/operator"
 	"github.com/perun-network/erdstall/tee"
 	"github.com/perun-network/erdstall/wire"
 )
@@ -34,6 +35,9 @@ type (
 		callbacks map[wire.ID]callback
 
 		subscription *Subscription
+
+		clientCfg operator.ClientConfig
+		hasConfig chan struct{}
 	}
 
 	// Subscription is returned by Subscribe() and can be used to iterate
@@ -64,6 +68,7 @@ func NewRPC(host string, port uint16) (*RPC, error) {
 	rpc := &RPC{
 		conn:      conn,
 		callbacks: make(map[wire.ID]callback),
+		hasConfig: make(chan struct{}),
 	}
 	if err := conn.SetReadDeadline(time.Time{}); err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


config/client.go
```diff
@@ -18,6 +18,8 @@ func ParseClientConfig() (cfg ClientConfig) {
 	flag.StringVar(&cfg.ChainURL, "chain-url", "ws://127.0.0.1:8545", "'protocol://ip:port'")
 	flag.StringVar(&cfg.OpHost, "op-host", "127.0.0.1", "IP of hostname")
 	flag.IntVar(&cfg.OpPort, "op-port", 8401, "Account derivation index.")
+	flag.StringVar(&cfg.OpHost, "op-host", "127.0.0.1", "IP/host name of operator")
+	flag.IntVar(&cfg.OpPort, "op-port", 8401, "Port of operator.")
```
sebastianst, line 25: The first commit adds there two duplicate lines but doesn't remove the old ones.

client/rpc.go
```diff
@@ -152,23 +165,33 @@ func (r *RPC) handleConnections() {
...
-				r.Log().Error("Received proof without subscription")
-				return
+		if !hasConfig {
+			if err := json.Unmarshal(data, &r.clientCfg); err != nil {
+				r.Log().Error("decoding client config: ", err)
+				continue
+			}
+			close(r.hasConfig)
+			hasConfig = true
+		} else {
```
sebastianst: I think it's easier to read if you add a `continue` in the first `if` branch and then put the normal path of execution after the `if`, instead of having it inside the `else` branch, which is always executed except for the very first msg.

gui/gui.go
```diff
@@ -187,7 +187,8 @@ func layout(g *gocui.Gui, client *client.Client) error {
 			return err
 		}
 		addr := client.Address().Hex()[:10]
-		v.Title = client.Config.UserName + fmt.Sprintf(" %s on %s", addr, client.Config.ChainURL)
+		v.Title = client.Config.UserName + fmt.Sprintf(" %s on %s", addr,
+			client.Config.ChainURL(client.OpConfig.NetworkID))
```
sebastianst: I think this expression is used several times and is worth being put into a method like `client.ChainURL() string`.

sebastianst (changes requested):

sebastianst (dismissed):

sebastianst (approved):

PR #93, opened 2021-01-27
Sync enclave

RmbRT:
* Made enclave synchronous, so that all commands are handled in one main goroutine.
* Removed lots of code.
* Reworked epoch and block processors.
* Split code into more files, grouped by topic.


tee/prototype/blockchain.go
```diff
@@ -5,84 +5,142 @@ package prototype
...
-		// first block
-		b.offset = blockNum
-		b.blocks = []*tee.Block{block}
-		return nil
+// valid successor block of the previous head. If the block is valid, returns
+// all of the block's deposit and exit events. If the block is invalid, the
+// chain remains unchanged.
+func (b *blockchain) PushVerify(
+	block *tee.Block,
+	params *tee.Parameters,
```
sebastianst, line 30: Since the parameters stay the same over the lifetime of a `blockchain`, we could consider putting a `*tee.Parameters` pointer into the `struct blockchain` so we don't need to pass it every time.

tee/prototype/blockchain.go
```diff
@@ -5,84 +5,142 @@ package prototype
...
+
+// verifySuccessorBlock verifies that next is a valid successor of head.
+func verifySuccessorBlock(next, head *tee.Block) error {
+	nextHash, headHash := next.Header().ParentHash, head.Hash()
+	nextN, headN := next.NumberU64(), head.NumberU64()
+
+	// TODO: Extend validation to test consensus (PoW).
+	//   This function then probably becomes a method on nextchain after we add a
+	//   consensus engine to nextchain.
+	switch true {
```
sebastianst:
```suggestion
	switch {
```

tee/prototype/blockchain.go
```diff
@@ -5,84 +5,142 @@ package prototype
...
+	}
 	return nil
 }
 
-// PruneUntil can be used to discard all blocks until the given block number
-func (b *blockchain) PruneUntil(blockNum uint64) {
-	if b.offset >= blockNum {
-		return
+// extractEvents extracts all deposit and exit events from a block.
+func (b *blockchain) extractEvents(
```
sebastianst: If I see correctly, functions `extractEvents` and `filterLogs` don't work on any fields of `blockchain`, so you might as well change them to free functions. Unless you put `params` into the struct... (at least for this method)

For my taste, methods should be methods on the primary object that they operate on. So in this case, either a method on `tee.Block` or if this doesn't make contextually sense, a free function. Note that parameter `epoch` is unused... something is wrong with our linter (or I am blind).

tee/prototype/blockchain.go
```diff
@@ -5,84 +5,142 @@ package prototype
...
+	return
+}
 
-	diff := blockNum - b.offset
-	b.blocks = b.blocks[diff:len(b.blocks)]
-	b.offset = blockNum
+// filterLogs partitions logs into different buckets of matching predicates.
+// Only logs from the Erdstall contract are filtered and other logs, as well as
+// those without a matching predicate, are discarded.
+func (b *blockchain) filterLogs(logs []*types.Log, preds []logPredicate, contract common.Address) [][]*types.Log {
```
sebastianst: free function

tee/prototype/enclave_command.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/perun-network/erdstall/tee"
+	log "github.com/sirupsen/logrus"
+)
+
+type (
+	// command represents all enclave command types.
+	command interface {
+		isACommand()
```
sebastianst: I'd just name it `command()`.

tee/prototype/enclave_command.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	return nil
+}
+
+// pushBlock pushes a new block onto the enclave's blockchain and updates the
+// enclave's state to reflect the new block. On error, the enclave remains
+// unchanged.
+func (e *Enclave) pushBlock(block *tee.Block) (phaseShift bool, _ error) {
+
```
sebastianst: remove \n

tee/prototype/enclave_command.go
```diff
@@ -0,0 +1,167 @@
...
+	// Cache any deposit proofs until the end of the phase.
+	e.depositProofCache = append(e.depositProofCache, e.generateDepositProofs(deps...)...)
+
+	e.epoch.ApplyDeposits(deps...)
+	e.epoch.RegisterExits(exits...)
+
+	phaseEnd := e.IsAtPhaseEnd()
+
+	return phaseEnd, nil
```
sebastianst:
```suggestion
	return e.IsAtPhaseEnd(), nil
```

sebastianst: I'd actually just let `pushBlock` return `error` and let the caller check for `e.IsAtPhaseEnd()`.

tee/prototype/enclave_command.go
```diff
@@ -0,0 +1,167 @@
...
+			LastBlock:     block.NumberU64(),
+			LastBlockHash: block.Hash(),
+			Accounts:      nil,
+		}
+	}
+
+	// Push block and if there is a phase shift, progress into the next epoch.
+	if phaseShift, err := e.pushBlock(block); err != nil {
+		return err
+	} else if phaseShift {
```
sebastianst: I'd actually just let `pushBlock` return `error` and check here directly for `e.IsAtPhaseEnd()`. Better decouplement of what `pushBlock` is doing and what the caller is doing.

tee/prototype/enclave.go
```diff
@@ -3,211 +3,196 @@
...
 	"github.com/ethereum/go-ethereum/common"
+
 	log "github.com/sirupsen/logrus"
-	perrors "perun.network/go-perun/pkg/errors"
-	"perun.network/go-perun/pkg/sync/atomic"
 
 	"github.com/perun-network/erdstall/tee"
+
+	perrors "perun.network/go-perun/pkg/errors"
+	"perun.network/go-perun/pkg/sync/atomic"
```
sebastianst: go-perun imports should be grouped together with other external imports like go-ethereum. I suggest you setup your IDE to call [`goimports`](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) at file save like `goimports -local github.com/perun-network/erdstall`.

tee/prototype/enclave.go
```diff
@@ -3,211 +3,196 @@
...
+		LastBlock     uint64      // The last known block height.
+		LastBlockHash common.Hash // The last known block's hash.
+
+		Accounts map[common.Address]*Acc // The users' accounts.
+	}
+
 	Enclave struct {
-		params  tee.Parameters
+		params *tee.Parameters // The enclave's parameters.
+		*State                 // The enclave's essential state.
```
sebastianst, line 31: Why as pointer?

tee/prototype/enclave.go
```diff
@@ -3,211 +3,196 @@
...
+// will continue receiving transactions and blocks until the last block of the
+// current phase is received via ProcessBlocks.
+//
+// The Enclave Interface methods will return an tee.ErrEnclaveStopped error after
+// the Enclave shut down.
+func (e *Enclave) Shutdown() {
+	log.Info("Enclave: shutting down when phase is finished")
+	select {
+	case e.commands <- &shutdownCmd{}:
+	case <-e.stopped:
```
sebastianst, line 192: I think it's fair to panic if `Shutdown` is called twice.

tee/prototype/enclave.go
```diff
@@ -3,211 +3,196 @@
...
-	// TODO: do we even want buffering? It may also be ok if ProcessX calls block.
-	bufSizeBlocks = 0 // incoming blocks buffer size
-	bufSizeTXs    = 0 // incoming tx buffer size
-	bufSizeProofs = 1 // proofs buffer size in #epochs
-)
+// EnclaveMaxCommandQueue is the number of commands that can be enqueued to the
+// enclave simultaneously. Once this number is surpassed, it is no longer
+// possible to ensure a FIFO execution of commands, and excess commands will
+// result in an error.
+const EnclaveMaxCommandQueue = 256 // for good measure.
```
sebastianst: why public?

tee/prototype/enclave.go
```diff
@@ -3,211 +3,196 @@
...
+	select {
+	case e.commands <- &processBlocksCmd{blocks: blocks, result: errCh}:
 		select {
-		case e.newBlocks <- blockReq{b, err}:
-		case <-e.done:
+		case errs := <-errCh:
+			errg := perrors.NewGatherer()
+			for _, e := range errs {
+				errg.Add(e)
+			}
```
sebastianst, line 145: we should add a constructor for this into `go-perun/pkg`...

tee/prototype/epoch.go
```diff
@@ -3,81 +3,198 @@
...
 	Epoch struct {
-		Number tee.Epoch
+		number tee.Epoch // Current deposit epoch.
+
+		exitLocked map[common.Address]struct{} // Withdrawing at end of epoch.
+		exitReqs   map[common.Address]struct{} // Requested exits.
+		balances   map[common.Address]*Acc     // Latest balances.
 
-		balances map[common.Address]*Bal
+		epochOutcome chan Outcome // The last epoch's outcome.
```
sebastianst:
```suggestion
		outcome chan Outcome // The last epoch's outcome.
```
what other outcome should it be as a field of `Epoch` ;)

tee/prototype/epoch.go
```diff
@@ -3,81 +3,198 @@
...
 type (
+	// Epoch manages deposits, transactions, and exits, as well as the
+	// progression of time in the system.
 	Epoch struct {
-		Number tee.Epoch
+		number tee.Epoch // Current deposit epoch.
+
+		exitLocked map[common.Address]struct{} // Withdrawing at end of epoch.
+		exitReqs   map[common.Address]struct{} // Requested exits.
+		balances   map[common.Address]*Acc     // Latest balances.
```
sebastianst: IMO the field name should be consistent with the type, so
```suggestion
		accs   map[common.Address]*Acc     // Latest account states.
```
or `accounts` or
```suggestion
		balances   map[common.Address]*Balance     // Latest balances.
```

tee/prototype/epoch.go
```diff
@@ -3,81 +3,198 @@
...
+	// Outcome contains all of an epoch's final balances, as well as all
+	// accounts that exited the system at the end of the epoch.
+	Outcome struct {
+		Epoch    tee.Epoch               // The finished epoch.
+		Exits    map[common.Address]*Acc // Exited accounts.
+		Accounts map[common.Address]*Acc // Final balances and nonces.
+	}
 )
 
+// NewEpoch creates an epoch manager that starts at the desired deposit epoch.
```
sebastianst: `manager` is a forbidden word ;)

sebastianst: We probably don't even need to have this function public...

tee/prototype/epoch.go
```diff
@@ -3,81 +3,198 @@
...
+// TxNum is the current transaction epoch number.
+func (e *Epoch) TxNum() tee.Epoch { return e.number - 1 }
+
+// ExitNum is the current exit epoch number.
+func (e *Epoch) ExitNum() tee.Epoch { return e.number - 2 }
+
+// Outcome has to be called exactly once after each phase shift and contains all
+// accounts that exited within this epoch, as well as all balances at the end of
+// the epoch. It blocks until the epoch in which it was last called ends.
+func (e *Epoch) Outcome() Outcome { return <-e.epochOutcome }
```
sebastianst: Is there any gain in exposing it as
```suggestion
func (e *Epoch) Outcome() <-chan Outcome { return e.epochOutcome }
```
so that, at the calling site, we can `select` on some other event like enclave closing?

tee/prototype/epoch.go
```diff
@@ -3,81 +3,198 @@
...
+	if _, locked := e.exitLocked[tx.Sender]; locked {
+		return errors.New("sender is locked for withdrawing")
+	} else if _, locked := e.exitLocked[tx.Recipient]; locked {
+		return errors.New("recipient is locked for withdrawing")
+	} else if tx.Nonce != sender.Nonce+1 {
+		return fmt.Errorf("nonce mismatch: %d != %d", tx.Nonce, sender.Nonce+1)
+	} else if sender.Value.Cmp((*big.Int)(tx.Amount)) < 0 {
+		return errors.New("insufficient balance")
+	} else if (*big.Int)(tx.Amount).Sign() < 0 {
+		return errors.New("negative balance")
```
sebastianst:
```suggestion
		return errors.New("negative amount")
```

tee/prototype/epoch.go
```diff
@@ -3,81 +3,198 @@
...
+
+	// Execute the transaction.
+	sender.Value.Sub(sender.Value, (*big.Int)(tx.Amount))
+	sender.Nonce = tx.Nonce
+
+	if recipient, ok := e.balances[tx.Recipient]; !ok {
+		e.balances[tx.Recipient] = &Acc{0, new(big.Int).Set((*big.Int)(tx.Amount))}
+	} else {
+		recipient.Value.Add(recipient.Value, (*big.Int)(tx.Amount))
 	}
```
sebastianst: I think there's code/logic duplication here and in `ApplyDeposits` that should be better encapsulated into a method

tee/test/generictest.go
```diff
@@ -159,17 +159,22 @@ func GenericEnclaveTest(t *testing.T, enc tee.Enclave) {
 		doWith(bp, alice.Exit, bob.Exit)
 	}
 
+	seal("exitPhase", 1)
+	bpsExit, err := enc.BalanceProofs()
+	requiree.NoError(err)
+	requiree.Len(bpsExit, 2, "users should still be in the system (frozen)")
```
sebastianst: "frozen" is misleading because we use this term in the case that the operator goes offline and the contract freezes. Do you mean something like "exit locked" here?

tee/test/generictest.go
```diff
@@ -159,17 +159,22 @@ func GenericEnclaveTest(t *testing.T, enc tee.Enclave) {
...
+	_, err = enc.DepositProofs()
+	requiree.NoError(err)
+
```
sebastianst, line 169: Also add an assert on length 0 for the deposit proofs.

sebastianst (changes requested):

tee/test/generictest.go
RmbRT: yes, good catch

tee/prototype/epoch.go
RmbRT: no, it is only called internally in the block processor after the epoch has already shifted, but I guess it would make sense to add an assertion that it really exists, and panic instead of deadlocking in case it does not exist. The enclave cannot close at that point because it only after the block processor is done with the epoch in which a shutdown was requested.

RmbRT: I added a log.Panic in case it is called more than once per epoch or before the epoch shift, because that would indicate a serious logic error.

tee/prototype/enclave_command.go
RmbRT: :+1:

tee/prototype/blockchain.go
RmbRT: removed epoch, but kept the functions as freestanding for now. I don't want to mess around in the rest of the code base outside the tee/prototype directory, as this would also make me rework of the whole tee.Block API. I'll open an issue for this.

tee/prototype/blockchain.go
RmbRT: I don't really like state duplication, I already feel bad for having the parameters inside the Enclave and also in Enclave.State. The blockchain doesn't really own the parameters, and I think therefore it shouldn't have them. it only needs the parameters to know the context in which it operates, but the parameters are not really part of what it is or does.

RmbRT: Same with the epoch. It needs it to verify the deposits and exits, but the epoch shouldn't be part of the blockchain's state.

tee/prototype/enclave.go
RmbRT: So that it is nil if the enclave is not yet fully initialised. It's just as a failsafe so that it panics on invalid usage.

tee/prototype/enclave.go
RmbRT: created issue 456.

tee/prototype/epoch.go
RmbRT: accs

sebastianst (approved):

